PR #1, opened 2016-07-16
Let cualloc decide which element types make sense

vchuravy:
The check in 6a1418 should be sufficient, but the check in the Array constructor is more stringent.

This is needed for the example at https://github.com/JuliaGPU/CUDAnative.jl/blob/3186b236c5ff329ea727318175bb39f7fe2a9800/examples/nvcxx/curand.jl to work.

### DIFF
--- a/src/arrays.jl
+++ b/src/arrays.jl
@@ -12,9 +12,7 @@ type CuArray{T,N} <: AbstractArray{T,N}
     len::Int
 
     function CuArray(::Type{T}, shape::NTuple{N,Int})
-        if !isbits(T)
-            throw(ArgumentError("CuArray with non-bit element type not supported"))
-        elseif (sizeof(T) == 0)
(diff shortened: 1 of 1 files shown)


maleadt:
Yeah, the check is too strict, but `sizeof(T)` doesn't cut it either:

```
julia> sizeof(AbstractString)
ERROR: argument is an abstract type; size is indeterminate
```

(FYI checking `DataType.abstract` doesn't prevent from `sizeof` failing either.)
I have this fixed in my local Array rework branch, will be committing tonight or tomorrow.

vchuravy:
Sounds good ;)

maleadt:
Oh I didn't notice I had pushed 6a14187077eac14204453e0e701cd656738311b3 already. Any objection to completely removing all checks from `CuArray` (including the broken `sizeof(T)` check) , letting `cualloc` doing all the checking?

PR #6, opened 2016-11-04
Support for copy(CuArray)

jmaebe:
Add support for copy(srcCuArray), i.e. copy!(dstCuArray,srcCuArray)


maleadt:
Changed the test, but other than that good to go!

PR #8, opened 2016-11-08
updated README with current repo name for install and test

waltsims:
this is something I stumbled upon while installing. quick fix.


maleadt:
Thanks!

PR #12, opened 2017-01-25
fix copy direction

adambrewster:
The test of copying memory to the GPU and back seems broken.


vchuravy:
Tested locally. Thanks!

PR #16, opened 2017-02-17
Correct error message and use `Pkg.dir`.

vchuravy:


maleadt:
Typo (`CUDAdrvi`)?
That said, I purposefully *did not* use `Pkg.dir` because you can also load packages using `JULIA_LOAD_PATH`, ie. outside of `JULIA_PKGDIR`...

maleadt:
See https://github.com/JuliaLang/julia/issues/8679

maleadt (approved):
Approved if it doesn't break CUDAnative.

PR #17, opened 2017-02-17
refactor stream constructor for CUDArt

vchuravy:


maleadt:
Why is this necessary? Especially after https://github.com/JuliaGPU/CUDArt.jl/pull/62?

src/stream.jl
```diff
@@ -9,6 +9,13 @@ const CuStream_t = Ptr{Void}
 type CuStream
     handle::CuStream_t
     ctx::CuContext
+    function CuStream(handle::CuStream_t, ctx=CuCurrentContext())
+        obj = new(handle, ctx)
+        block_finalizer(obj, ctx)
+        finalizer(obj, finalize)
+        return obj
+    end
+    CuStream(handle::CuStream_t, ctx::CuContext, _) = new(handle, ctx)
```
maleadt, line 18: Why is this necessary? I don't really like having such a three-argument ctor where the third argument is ignored...

maleadt (changes requested):

vchuravy:
We have no constructor that just takes a handle: https://github.com/JuliaGPU/CUDArt.jl/blob/a6cf291d09adee29d4a36ec8742c4c4e947ffd13/src/stream.jl#L16

src/stream.jl
vchuravy: I originally wanted to define:
```
CuStream() = new(convert(CuStream_t, C_NULL), C_NULL)
```

but that is ambiguous with the `flags::Integer=0`

PR #19, opened 2017-02-27
Primary contexts for CUDArt

vchuravy:
This adds the necessary infrastructure to handle primary contexts improving interaction between CUDArt and CUDAdrv.


vchuravy:
Rebased onto current master and added tests.

vchuravy:
Once we have reset a context or destroyed it we should no longer free any resources attached to it since they will be destroyed already by the driver. In the case of primary contexts the story is a bit more complicated, because a reset followed by a retain will give you a valid context at the same address, so the driver will not throw an error (since there is a valid context) but will seqfault if you free a previous attached resource....

@maleadt if you are ok with this approach I will extend this to the other managed resources.

maleadt:
Right. This would also protect against ouf-of-order finalizers (where contexts are finalized before their resources), making `can_finalize` redundant.
OTOH, it puts the burden on downstream objects to check if their linked context is still valid...

vchuravy:
Yeah I was hoping that there would be an API to query if a context is still valid, but even that wouldn't work in the primary context case... I am not certain if we can replace can_finalize with this but it might be worth a try.

Should we fold owned and isvalid into a single bitmask field?

maleadt:
> Should we fold owned and isvalid into a single bitmask field?

Nah, that's micro-optimization for an object that's pretty heavy to allocate anyway. A more sensible optimization would be to cache `CuCurrentContext`, but that's for another time and PR.

vchuravy:
~The test failure only happens if I run julia with `--compilecache=yes` ~ Silly me.

vchuravy:
CUDAnative test pass locally.

vchuravy:
squashed commits, CUDAnative and CUDAdrv both pass tests.

maleadt:
Ha, you went quite a bit further than I intended, ripping out _all_ of `gc.jl` instead of just the out-of-order finalization check (the `block` methods were intended to _force_ in-order finalization). But, on the other hand, why not... Even if we trigger more out-of-order finalizer runs that way, they'll all be valid and tearing down the context takes care of destroying everything anyway.

vchuravy:
Oh yeah... I just thought that keeping two systems that do the same thing
is kinda pointless ;)

vchuravy:
Rebase onto master again, hopefully I adapted to the test changes correctly. Tests pass locally.

src/module.jl
```diff
@@ -18,7 +18,7 @@ type CuModule
     If the Julia debug level is 2 or higher (or, on 0.5, if CUDAdrv is loaded in DEBUG
     mode), line number and debug information will be requested when loading the PTX code.
     """
-    function CuModule(data, options::Dict{CUjit_option,Any}=Dict{CUjit_option,Any}())
+    function CuModule(data, options::Dict{CUjit_option,Any}=Dict{CUjit_option,Any}(); ctx = CuCurrentContext())
```
maleadt: Why do you do this? We do an API call to load the module data in this ctor, so it can't be that the context is anything else than the default one?

src/module.jl
```diff
@@ -72,7 +73,7 @@ Create a CUDA module from a file containing PTX code.
 Note that for improved error reporting, this does not rely on the corresponding CUDA driver
 call, but opens and reads the file from within Julia instead.
 """
-CuModuleFile(path) = CuModule(open(readstring, path))
+CuModuleFile(path, ctx = CuCurrentContext()) = CuModule(open(readstring, path); ctx = ctx)
```
maleadt: idem

src/util/logging.jl
```diff
@@ -26,7 +26,7 @@ is set."
 end
 @inline trace(msg...; kwargs...) = trace(STDERR, msg...; kwargs...)
 macro trace(args...)
-    TRACE && return esc(Expr(:call, :trace, args...))
+    TRACE && return esc(Expr(:call, :(CUDAdrv.trace), args...))
```
maleadt: Nice catch. Might be better to keep the `esc` fine grained though: leave the `trace` unescaped, and pass `$(map(esc, args)...)`

src/util/logging.jl
```diff
@@ -39,7 +39,7 @@ variable is set."
 end
 @inline debug(msg...; kwargs...) = debug(STDERR, msg...; kwargs...)
 macro debug(args...)
-    DEBUG && return esc(Expr(:call, :trace, args...))
+    DEBUG && return esc(Expr(:call, :(CUDAdrv.trace), args...))
```
maleadt: idem

maleadt (changes requested):
Great, thanks! Some minor nitpicks. Merge/tag at your will.

PR #23, opened 2017-04-04
make double free possible

SimonDanisch:
Julias GC doesn't know about memory shortage on the GPU, which is why sometimes a manual free is needed. To implement that, we need to allow double frees, since we still have the finalizer!

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -37,6 +37,7 @@ function finalize(a::CuArray)
     if isvalid(a.devptr.ctx)
         @trace("Finalizing CuArray at $(Base.pointer_from_objref(a))")
         Mem.free(a.devptr)
+        a.devptr = CU_NULL
     else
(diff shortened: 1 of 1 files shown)


maleadt:
Why would the finalizer run twice? Recalling this Slack excerpt:

```
stefan [3:18 PM] ￼
so if I do this:
    finalizer(x, fin)
    …
    finalize(x)
then when x is GC’d no finalizer will be called on x?

yuyichao [3:18 PM] ￼
correct
You can obviously call the finalizer multiple times and no one can stop that.
if you call the finalizer directly then there's no way the gc can know that it's called already
```

... or are we making this impossible by calling our finalizer `finalize`, hence when calling `finalize(CuArray)` we're calling our finalizer directly instead of calling into the GC? That would be nasty!

SimonDanisch:
Oh, I didn't realize that you can call finalize that way.. That's the way to go then I suppose!

PR #25, opened 2017-04-05
Update library.jl

musm:
Minor fixes for windows, still doesn't pass since the script looks for gcc

### DIFF
--- a/examples/compilation/library.jl
+++ b/examples/compilation/library.jl
@@ -70,14 +70,18 @@ const toolchain = Ref{Toolchain}()
 function discover_toolchain()
     # Check availability NVCC
     if haskey(ENV, "NVCC")
-        nvcc = get(ENV, "NVCC")
-    else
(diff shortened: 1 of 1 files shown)


maleadt:
Is `nvcc` somewhat smarter wrt. detecting the host compiler on Windows than on Linux, or does it just pick whichever `cl.exe` it finds on the PATH? The reason I added that `gcc` detection logic is because `nvcc` does exactly that on Linux: pick the first `gcc` binary it can find, disregarding version (in)compatibilities. Hence my version checking, and specifying the correct one using `-ccbin`.
If it is, just guard all that `gcc`-specific logic behind an `is_linux()`. If not, we might need to port it. Or just mark this code `!is_windows()`, it's an example after all...

musm:
I think it just picks whichever `cl.exe` it find in path (set up by running vcvarsall). The library.jl file needs to be updated to run `vcvarsall.bat` which set's up the c/c++ compiler.

PR #29, opened 2017-04-06
Fix windows search for nvcc and compiler

musm:
This PR and the latest pushes to master now fix `Pkg.test` for me

supersedes https://github.com/JuliaGPU/CUDAdrv.jl/pull/25 and https://github.com/JuliaGPU/CUDAdrv.jl/pull/24

### DIFF
--- a/examples/compilation/library.jl
+++ b/examples/compilation/library.jl
@@ -70,14 +70,18 @@ const toolchain = Ref{Toolchain}()
 function discover_toolchain()
     # Check availability NVCC
     if haskey(ENV, "NVCC")
-        nvcc = get(ENV, "NVCC")
-    else
+        nvcc = ENV["NVCC"]
+    elseif haskey(ENV, "CUDA_PATH")
+        nvcc = joinpath(ENV["CUDA_PATH"], "bin", "nvcc") * (is_windows() ? ".exe" : "")
+    elseif !is_windows()
         try
             nvcc = chomp(readstring(pipeline(`which nvcc`, stderr=DevNull)))
         catch ex
             isa(ex, ErrorException) || rethrow(ex)
-            rethrow(ErrorException("could not find NVCC -- consider specifying with NVCC environment variable"))
+            rethrow(ErrorException("could not find NVCC; consider setting the NVCC environment variable or the CUDA_PATH environment variable"))
         end
+    else
+        throw(ErrorException("could not find NVCC; consider setting the NVCC environment variable or the CUDA_PATH environment variable"))
     end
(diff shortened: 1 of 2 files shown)


examples/compilation/library.jl
```diff
@@ -70,14 +70,18 @@ const toolchain = Ref{Toolchain}()
 function discover_toolchain()
     # Check availability NVCC
     if haskey(ENV, "NVCC")
-        nvcc = get(ENV, "NVCC")
-    else
+        nvcc = ENV["NVCC"]
+    elseif haskey(ENV, "CUDA_PATH")
```
musm, line 74: On windows, a default installation of CUDA will install this environment variable, so it's good to include it here for automatic detection at least.

examples/compilation/library.jl
```diff
@@ -91,67 +95,78 @@ function discover_toolchain()
...
+            error("could not find a suitable host compiler (your NVCC $version needs GCC <= $(get(hostcc_maxver)))")
+        end
+        sort!(hostcc_possibilities; rev=true, lt=(a, b) -> a[2]<b[2])
+        hostcc = hostcc_possibilities[1]
+
+        push!(flags, hostcc[1])
+    else
+        vc_versions = ["VS140COMNTOOLS", "VS120COMNTOOLS", "VS110COMNTOOLS", "VS100COMNTOOLS"]
+        vs_cmd_tools_dir = ENV[vc_versions[first(find(x -> haskey(ENV, x), vc_versions))]]
+        hostccbin = joinpath(dirname(vs_cmd_tools_dir), "..", "..", "VC", Sys.WORD_SIZE == 64 ? "amd64" : "", "cl.exe")
```
musm, line 164: If I don't include `cl.exe`  it has troubles compiling (perhaps it chooses the wrong compiler or something...?).
Another option would be to somehow run `vsvarsall.bat`  and then compile with `nvcc` but I don't know how one could do that in julia; typically in a cmd prompt session you run `vsvarsall.bat amd64 (or another compilation option) and then `nvcc` runs without having to set `--compiler-bindir` .

maleadt: You could 'source' `vcvarsall.bat` every time before running `nvcc`, but that would be slooow. So lets just leave it at this, if that works for you, until someone finds (the need for) a better fix.

maleadt:
Looks good, thanks for working on this!
I'll probably go and make the errors reported by this example non-fatal though, as CUDAdrv.jl shouldn't require the full CUDA toolkit be installed (only the driver library), but nonetheless it's good to have it compatible with Windows.

musm:
great thanks

PR #30, opened 2017-04-08
Update CUDAdrv.jl

musm:


src/CUDAdrv.jl
```diff
@@ -3,21 +3,24 @@ __precompile__()
 module CUDAdrv
 
 using Compat
-import Compat.String
+using Compat.String
```
maleadt, line 6: Why this change? The [Compat.jl README](https://github.com/JuliaLang/Compat.jl#usage) suggests otherwise.

maleadt (approved):

src/CUDAdrv.jl
musm: Doesn't look like the method is  extended in the repo, so changed it to `using`  for safety. Not  essential, though.

https://docs.julialang.org/en/stable/manual/modules/
"The import keyword supports all the same syntax as using, but only operates on a single name at a time. It does not add modules to be searched the way using does. import also differs from using in that functions must be imported using import to be extended with new methods."

vchuravy: `import` is the canonical way of only bringing only one name into the current namespace, so I would also prefer the old version.

src/CUDAdrv.jl
```diff
@@ -3,21 +3,24 @@ __precompile__()
...
+const ext = joinpath(dirname(@__DIR__), "deps", "ext.jl")
+if isfile(ext)
+    include(ext)
+else
+    error("Unable to load dependency file $ext.\nPlease run Pkg.build(\"CUDAdrv\") and restart Julia.")
+end
 const libcuda = libcuda_path
 
-include("util/logging.jl")
+include(joinpath("util", "logging.jl"))
```
vchuravy, line 16: Are these change really necessary under Windows? I understand the intention behind them, but I find it a bit odd.

vchuravy (commented):
Thank you for your contributions. Just as a side note, a lot of these changes seem to be primarily style/preference based, which makes it hard for reviewers to find the logic and behaviour changes and the commits should be split into logic changes vs style changes.

src/CUDAdrv.jl
musm: Backtraces get messes up if you use forward slash on Windows, it still prints but will have forward and backslashes in the path, so I think good practice to always use join path

src/CUDAdrv.jl
musm: Sure, base uses this too.

src/CUDAdrv.jl
vchuravy: Okay that makes sense to me.

vchuravy (approved):
Please squash your commits and the LGTM.

src/CUDAdrv.jl
musm: I just double checked this. For example 
```julia
module A
include("b/b.jl")
end
````
produces
```julia
julia> include("C:\\Users\\Mus\\Desktop\\testt\\A.jl")
julia> using A
julia> A.foo()
ERROR: MethodError: no method matching foo()
Closest candidates are:
  foo(::Any) at C:\Users\Mus\Desktop\testt\b/b.jl:1
```
note the weird forward slash at the end

vs.

```julia
module A
include(joinpath("b", "b.jl"))
end
````
produces
```julia
julia> include("C:\\Users\\Mus\\Desktop\\testt\\A.jl")
julia> using A
julia> A.foo()
ERROR: MethodError: no method matching foo()
Closest candidates are:
  foo(::Any) at C:\Users\Mus\Desktop\testt\b\b.jl:1
```
file path has correct backslash

I believe there has been some discussion regarding a more sophisticated filepaths abstraction a la https://github.com/JuliaLang/Juleps/pull/26 etc.  It would be nice if this was not necessary.

src/CUDAdrv.jl
musm: So, I think in this case it doesn't actually matter too much since we already do `using Compat` so all the names are available anyways. The `using Compat.String` makes `String` available but not extendable, so it's safer in case we accidentally try to add a method to `String` (pedantic and unlikely, but I believe this is the distinction).

If on the other hand, we did not have `using Compat` and just wanted to use Compat's String method then `import Compat.String` is the correct/only way to bring into scope just that binding.

Please correct me if I'm wrong here.

tkelman: > `import` is the canonical way of only bringing only one name into the current namespace, so I would also prefer the old version.

Only for bringing a module name into the namespace without its exports. `import Mod` vs `using Mod` is somewhat backwards compared to `import Mod.func` vs `using Mod.func`. For functions, `import` allows extension and `using` doesn't, so `using` is safer.

musm: Using `using Test.func` on individual names *does not*  add the exported names of the `Test` module to be searched  as in `using Test`, but acts like `import Test.func`  but without allowing accidental function extensions. Perhaps this point should be better clarified in the docs.

BTW why is it even allowed to `import Test` a whole module? Perhaps that should be tightened up.

E.g.

```julia

julia> module Test

       export foo,goo

       foo() = 1
       goo() = 2
       hoo() = 3
       end
Test

julia> using Test.hoo

julia> hoo(x) = x
ERROR: error in method definition: function Test.hoo must be explicitly imported to be extended

julia> foo()
ERROR: UndefVarError: foo not defined

julia> goo()
ERROR: UndefVarError: goo not defined

julia> hoo()
3

## restart and try 


julia> import Test.hoo

julia> hoo(x) = x
hoo (generic function with 2 methods)

julia> foo()
ERROR: UndefVarError: foo not defined

julia> goo()
ERROR: UndefVarError: goo not defined

julia> hoo()
3
```

tkelman: > why is it even allowed to `import Test` a whole module?

It only brings in the module name. `importall Test` imports all of its functions (I forget whether just the exported ones, or absolutely everything) for extension, and is sketchy enough that there's rarely a good reason to use it.

PR #31, opened 2017-04-10
Clean up devices and add show method

musm:
Is the show method causing stack overflows?

### DIFF
--- a/src/devices.jl
+++ b/src/devices.jl
@@ -12,7 +12,7 @@ const CuDevice_t = Cint
 function devcount()
     count_ref = Ref{Cint}()
     @apicall(:cuDeviceGetCount, (Ptr{Cint},), count_ref)
-    return count_ref[]
+    return convert(Int, count_ref[])
 end
 
 immutable CuDevice
@@ -27,7 +27,14 @@ immutable CuDevice
     end
 end
 
-Base.convert(::Type{CuDevice_t}, dev::CuDevice) = dev.handle
+function Base.show(io::IO, dev::CuDevice)
+    nam = name(dev)
+    tmem = round(totalmem(dev)/(1024^2))
+    cap = capability(dev)
+    print(io, "device[$(Int(dev.ordinal))]: $(nam), capability $(cap.major).$(cap.minor), total memory= $tmem MB")
+end
+
+Base.convert{T<:Integer}(::Type{T}, dev::CuDevice) = convert(T, dev.handle)
 
 Base.:(==)(a::CuDevice, b::CuDevice) = a.handle == b.handle
 Base.hash(dev::CuDevice, h::UInt) = hash(dev.handle, h)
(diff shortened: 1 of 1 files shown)


src/devices.jl
```diff
@@ -27,7 +27,14 @@ immutable CuDevice
...
 
-Base.convert(::Type{CuDevice_t}, dev::CuDevice) = dev.handle
+function Base.show(io::IO, dev::CuDevice)
+    nam = name(dev)
+    tmem = round(totalmem(dev)/(1024^2))
+    cap = capability(dev)
+    print(io, "device[$(Int(dev.ordinal))]: $(nam), capability $(cap.major).$(cap.minor), total memory= $tmem MB")
+end
+
+Base.convert{T<:Integer}(::Type{T}, dev::CuDevice) = convert(T, dev.handle)
```
vchuravy, line 37: Why is this change necessary?

src/devices.jl
```diff
@@ -46,7 +53,7 @@ end
 function totalmem(dev::CuDevice)
     mem_ref = Ref{Csize_t}()
     @apicall(:cuDeviceTotalMem, (Ptr{Csize_t}, CuDevice_t), mem_ref, dev)
-    return mem_ref[]
+    return convert(Int, mem_ref[])
```
vchuravy, line 56: Why do you need to convert this from unsigned to signed? This will be problematic on 32bit systems if `totalmem` exceeds `typemax(Int32)` (which if my math is right would be more than 2GB)

src/devices.jl
```diff
@@ -171,10 +178,10 @@ function list_devices()
     for i = 0:cnt-1
         dev = CuDevice(i)
         nam = name(dev)
-        tmem = round(Integer, totalmem(dev) / (1024^2))
+        tmem = round(totalmem(dev)/(1024^2))
```
vchuravy, line 181: Please keep the spaces around the `/`.

vchuravy (changes requested):

src/devices.jl
```diff
@@ -27,7 +27,14 @@ immutable CuDevice
     end
 end
 
-Base.convert(::Type{CuDevice_t}, dev::CuDevice) = dev.handle
+function Base.show(io::IO, dev::CuDevice)
+    nam = name(dev)
```
vchuravy, line 31: Under a debug build this will cause a stack overflow since `@trace` will call show on the arguments of the api call. And `dev` is part of the apicall for name https://github.com/JuliaGPU/CUDAdrv.jl/blob/45c9ed0254817ba7dd322309ec4c01704debe0ca/src/devices.jl#L40

musm: I see, thanks

src/devices.jl
musm: Was hoping to make things more user friendly, it's a lot easier to read  as an Int  as opposed toa UInt c.f.
```julia
julia> totalmem(CuDevice(0))
0x0000000080000000

#vs  
julia> totalmem(CuDevice(0))
2147483648
```

musm:
ganna close since I don't think this was successful

PR #35, opened 2017-05-25
Add more detail to a couple of docstrings

timholy:


maleadt:
Yeah, documentation is really lacking with my packages... Thanks!

PR #44, opened 2017-06-23
Add launch() method with default arguments and constructor to convert CudaArray to CuArray.

mikhail-j:
I added a new constructor for CUDAdrv.CuArray to convert CUDArt.CudaArray to CUDAdrv.CuArray.

A separate launch() method (shmem=0 and stream=CuDefaultStream()) was included in "execution.jl".

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -45,13 +45,6 @@ CuArray
         # semi-hidden constructor, only called by unsafe_convert
         new{T,N}(devptr, shape)
     end
-    # Specifically convert existing CUDArt CudaArray to CuArray
-    function (::Type{CuArray{T,N}}){T,N}(rt_array::CudaArray{T, N})
-        local devptr = DevicePtr{T}(rt_array.ptr.ptr, rt_array.ptr.ctx)
-
-        # This should be carefully freed (both CudaArray and CuArray will share the same memory address!)
-        return new{T,N}(devptr, rt_array.dims)
-    end
 end
 
 @compat const CuVector{T} = CuArray{T,1}
--- a/src/array.jl
+++ b/src/array.jl
@@ -45,6 +45,13 @@ CuArray
         # semi-hidden constructor, only called by unsafe_convert
         new{T,N}(devptr, shape)
     end
+    # Specifically convert existing CUDArt CudaArray to CuArray
+    function (::Type{CuArray{T,N}}){T,N}(rt_array::CudaArray{T, N})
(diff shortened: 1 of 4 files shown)


src/execution.jl
```diff
@@ -56,6 +73,14 @@ This is a low-level call, prefer to use [`cudacall`](@ref) instead.
     _launch(f, griddim, blockdim, shmem, stream, args)
 end
 
+@inline function launch{N}(f::CuFunction, griddim::CuDim3, blockdim::CuDim3,
```
maleadt: Just make this a 1-line definition calling `launch`, don't duplicate its body to call `_launch`.
Also add the second calling syntax to the documentation above, and mention the default values for those arguments.

src/array.jl
```diff
@@ -45,6 +45,13 @@ CuArray
         # semi-hidden constructor, only called by unsafe_convert
         new{T,N}(devptr, shape)
     end
+    # Specifically convert existing CUDArt CudaArray to CuArray
```
maleadt: No, this doesn't belong here. Add it to CUDArt, which already depends on CUDAdrv.

src/execution.jl
```diff
@@ -1,3 +1,20 @@
+
+# CUDArt defines CudaArray
```
maleadt: Idem. If adding to CUDArt, you don't need those `isdefined` checks.

Also, we don't tend to use parentheses around the if conditions.

maleadt (changes requested):

maleadt:
Also, tests.

src/execution.jl
maleadt: Also, maybe make that new `launch` accept `kwargs`, then it would mimic `cudacall` which has both a `kwargs` version, and a non-`kwargs` one for performance.

mikhail-j:
I have added tests and updated the documentation for launch() has been updated with default values for 'shmem' and 'stream'.

I checked the error in [Julia master build](http://ci.maleadt.net/buildbot/julia/builders/CUDAdrv.jl%20CI%3A%20Julia%20master%20%28x86-64%29/builds/48) is related to module.jl, which I did not modify.

maleadt:
Thanks for your suggestions on this. I proposed a similar version, using `kwargs` instead, and accepting `Cudim` all over the place (instead of locally in `launch`). This should enable your use of `launch`, but fix it in a more thorough manner.

PR #46, opened 2017-06-24
WIP: debug 0.7 failure

maleadt:
Started happening after 6db340cab4695d0c58e5c1126df3e0a7a6138da8, only reproducible on CI.


maleadt:
Gah, printing the data removes the issue. Would this really be GC corruption?

PR #48, opened 2017-06-30
Fail gracefully when no CUDA available

MikeInnes:
This patch makes the package installable and loadable on machines without CUDA. This is desirable so that downstream packages don't have to do conditional loading hacks to get CPU-only setups.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -51,7 +51,7 @@ function find_libcuda()
       warn("CUDA driver library cannot be found (specify the path to $(libcuda_name) using the CUDA_DRIVER environment variable).")
       open(ext, "w") do fh
           write(fh, """
-              ENV["ONLY_LOAD"] = "true"
+              ENV["CUDADRV_ONLY_LOAD"] = "true"
               """)
       end
       exit()
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -47,7 +47,15 @@ function find_libcuda()
             end
         end
     libcuda = Libdl.find_library(libcuda_name, libcuda_locations)
-    isempty(libcuda) && error("CUDA driver library cannot be found (specify the path to $(libcuda_name) using the CUDA_DRIVER environment variable).")
+    if isempty(libcuda)
+      warn("CUDA driver library cannot be found (specify the path to $(libcuda_name) using the CUDA_DRIVER environment variable).")
+      open(ext, "w") do fh
+          write(fh, """
+              ENV["ONLY_LOAD"] = "true"
+              """)
+      end
+      exit()
+    end
 
     # find the full path of the library
     # NOTE: we could just as well use the result of `find_library,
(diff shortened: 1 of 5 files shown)


deps/build.jl
```diff
@@ -47,7 +47,15 @@ function find_libcuda()
             end
         end
     libcuda = Libdl.find_library(libcuda_name, libcuda_locations)
-    isempty(libcuda) && error("CUDA driver library cannot be found (specify the path to $(libcuda_name) using the CUDA_DRIVER environment variable).")
+    if isempty(libcuda)
+      warn("CUDA driver library cannot be found (specify the path to $(libcuda_name) using the CUDA_DRIVER environment variable).")
+      open(ext, "w") do fh
+          write(fh, """
+              ENV["CUDADRV_ONLY_LOAD"] = "true"
```
vchuravy, line 54: I would prefer having a Julia variable being set instead of round tripping through the environment.

maleadt:
What a coincidence, I've been working on almost exactly the same thing for the past couple of hours (on a plane, offline)... I was getting rid of the magical env var though, which I think is not really nice. Let me push to github.

EDIT: WIP over at https://github.com/JuliaGPU/CUDAdrv.jl/tree/tb/stubs
I decided to still require _a_ installation of CUDA, but allow usage of the stub libraries instead (this could still be changed). After some redesign of inner functionality it is even possible to run some of the tests.
I'll be spending more time on this early next week.

maleadt:
@MikeInnes: does https://github.com/JuliaGPU/CUDAdrv.jl/commit/171b89864ea6c1e207a7099313139d5fd0dc389f work for you as well?
Also doing https://github.com/maleadt/LLVM.jl/commit/39b3c44431926dff19fc96f4b7e02b8545b7db1a to enable https://github.com/JuliaGPU/CUDAnative.jl/commit/aa14d7a2abd5ce02c6f11492737ac8061d3fefd2 (@SimonDanisch would that simplify conditional module usage for GPUArrays?)

src/CUDAdrv.jl
```diff
@@ -5,17 +5,15 @@ module CUDAdrv
 using Compat
 using Compat.String
 
-const ext = joinpath(dirname(@__DIR__), "deps", "ext.jl")
-if isfile(ext)
-    include(ext)
-elseif haskey(ENV, "ONLY_LOAD")
+include("../deps/ext.jl")
```
maleadt, line 8: just FYI, the `joinpath` is for better backtraces on Windows.

maleadt:
OK, I'm going to assume #49 accomplishes the same goal: being able to load CUDAdrv if the build failed. If there's remaining issues with that functionality, just open an issue.

src/CUDAdrv.jl
MikeInnes: Hmm, Julia should really be normalising these in some way so that's not necessary

maleadt: Agreed, but Base seems to be moving in the other direction (eg. JuliaLang/julia#21405).

MikeInnes: Ouch

PR #49, opened 2017-07-01
Support for loading/testing the package if the build failed.

maleadt:
Alternative to #48:
- supports loading/testing the package if the build failed (ie no `ext.jl`), only failing upon actual API calls
- ditches `ONLY_LOAD`/`CUDADRV_ONLY_LOAD` flags

Also:
- supports running tests without Documenter
- eagerly decodes some of the build error codes (to avoid user confusion)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ notifications:
 env:
   global:
     - DOCUMENTER_DEBUG=true
-    - ONLY_LOAD=true
 
 after_success:
   - julia -e 'Pkg.add("Documenter")'
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -14,14 +14,14 @@ macro apicall(libpath, fn, types, args...)
         lib = Libdl.dlopen($(esc(libpath)))
         sym = Libdl.dlsym(lib, $(esc(fn)))
 
-        status = ccall(sym, Cint, $(esc(types)), $(map(esc, args)...))
-        status != 0 && error("CUDA error $status calling ", $fn)
+        ccall(sym, Cint, $(esc(types)), $(map(esc, args)...))
     end
 end
 
 function version(libpath)
     ref = Ref{Cint}()
-    @apicall(libpath, :cuDriverGetVersion, (Ptr{Cint}, ), ref)
+    status = @apicall(libpath, :cuDriverGetVersion, (Ptr{Cint}, ), ref)
+    @assert status == 0
     return VersionNumber(ref[] ÷ 1000, mod(ref[], 100) ÷ 10)
 end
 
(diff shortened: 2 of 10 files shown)


MikeInnes:
This solves the basic issue, but gives output like this:

```julia
julia> Pkg.build("CUDAdrv")
INFO: Building CUDAdrv
===============================[ ERROR: CUDAdrv ]===============================

LoadError: Could not find the CUDA driver library (specify the path to libcuda using the CUDA_DRIVER environment variable).
while loading /Users/mike/.julia.1/v0.6/CUDAdrv/deps/build.jl, in expression starting on line 119

================================================================================

================================[ BUILD ERRORS ]================================

WARNING: CUDAdrv had build errors.

 - packages with build errors remain installed in /Users/mike/.julia.1/v0.6
 - build the package(s) and all dependencies with `Pkg.build("CUDAdrv")`
 - build a single package by running its `deps/build.jl` script

================================================================================

julia> using CUDAdrv
WARNING: CUDAdrv.jl has not been configured, and will not work properly.
WARNING: Please run Pkg.build("CUDAdrv") and restart Julia.
```

I suggest just having a warning at build time and/or throwing errors on use of CUDAdrv rather than on load. Otherwise downstream GPUArrays users are going to think everything is broken, and you're going to get a lot of issues from people who ran `Pkg.build` over and over again.

maleadt:
Yes, that output is as intended: I don't want to support loading the package when the build failed, without any warning whatsoever. I'm open to making the warnings opt-out though, maybe an `SILENT` env var?

MikeInnes:
There can still be a warning/error, just delayed a bit longer.

If CUDAdrv looks like it failed to install then we're back to where we started as far as downstream dependencies are concerned.

maleadt:
Not sure I'm following, wouldn't it be enough for downstream packages to be able to import CUDAdrv without warnings and inspect (and alter their behaviour) based on the newly-added `CUDAdrv.configured` flag indicating whether the build succeeded?

MikeInnes:
That's all fine until users stop Julia and run `Pkg.build()` again based on what they've been told. If they even get that far; many will give up after seeing big blocks of red in their terminal instead of a clean install.

An ENV var is messy for the load warning but gets the job done. For the build we do of course want to let people know that CUDA isn't set up, but a simple `warn` would achieve that; otherwise we tell everyone without CUDA that their GPUArrays installation is broken.

maleadt:
> That's all fine until users stop Julia and run Pkg.build() again based on what they've been told.

I'd hide that warning too if eg. GPUArrays imports CUDAdrv/CUDAnative with that hypothetical `SILENT` env var set.

> For the build we do of course want to let people know that CUDA isn't set up, but a simple warn would achieve that; otherwise we tell everyone without CUDA that their GPUArrays installation is broken.

Makes sense from the GPUArrays POV, but not so much for eg. CUDAnative users. In that case you'd _want_ the failure to be all red and emphasized, because it'll break the very thing that that the user was trying to install. In other words, there's a balance to be struck here, and I don't want to "sacrifice" that usability just because the language is missing a feature (conditional dependencies).

PR #51, opened 2017-07-27
Rework and internalize DevicePtr

maleadt:
This PR renames `DevicePtr` to `OwnedPtr`, and makes it an internal implementation detail. This impacts `cudacall` invocations, which now need to use `Ptr` instead of `DevicePtr`, because this actually matches the C type signature of kernel functions (as well as being easier to use, cfr. `ccall`). As such, this'll probably break some downstream packages (most notably CUDArt).


maleadt:
cc @SimonDanisch: you use DevicePtr in GPUArrays, and will need to switch to regular Ptr

PR #52, opened 2017-07-31
better pointer handling + refcounting

SimonDanisch:
Track device pointer better, to not double free when creating multiple CuArrays from the same shared pointer.
Also introduces some pointer arithmetic on owned pointer!

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -33,17 +33,19 @@ CuArray
         elseif (sizeof(T) == 0)
             throw(ArgumentError("CuArray with zero-sized element types does not make sense"))
         end
-
         len = prod(shape)
         ptr = Mem.alloc(T, len)
-
+        Mem.retain(ptr)
         obj = new{T,N}(ptr, shape)
         finalizer(obj, unsafe_free!)
         return obj
     end
     function (::Type{CuArray{T,N}}){T,N}(shape::NTuple{N,Int}, ptr::OwnedPtr{T})
         # semi-hidden constructor, only called by unsafe_convert
-        new{T,N}(ptr, shape)
+        obj = new{T, N}(ptr, shape)
+        Mem.retain(ptr)
+        finalizer(obj, unsafe_free!)
+        obj
     end
 end
 
@@ -52,11 +54,13 @@ end
 
 function unsafe_free!(a::CuArray)
     ptr = pointer(a)
-    if isvalid(ptr.ctx)
+    if !isvalid(ptr.ctx)
+        @trace("Skipping finalizer for CuArray object at $(Base.pointer_from_objref(a))) because context is no longer valid")
+    elseif !Mem.release(ptr)
+        @trace("Skipping finalizer for CuArray object at $(Base.pointer_from_objref(a))) because pointer is held by another object")
+    else
         @trace("Finalizing CuArray object at $(Base.pointer_from_objref(a))")
         Mem.free(ptr)
-    else
-        @trace("Skipping finalizer for CuArray object at $(Base.pointer_from_objref(a))) because context is no longer valid")
     end
 end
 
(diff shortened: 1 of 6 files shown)


SimonDanisch:
woops, pushed the gitignore change to master ... Guess it should be okay ;)
But I definitely should squash this!

src/array.jl
```diff
@@ -54,7 +58,10 @@ function unsafe_free!(a::CuArray)
     ptr = pointer(a)
     if isvalid(ptr.ctx)
         @trace("Finalizing CuArray object at $(Base.pointer_from_objref(a))")
-        Mem.free(ptr)
+        if Mem.refcount(ptr) == 1
```
maleadt: why not put this in `Mem.free`? Then you only need an additional `Mem.retain` to be called from the `CuArray` constructors.

src/pointer.jl
```diff
@@ -13,12 +13,19 @@ function Base.:(==)(a::OwnedPtr, b::OwnedPtr)
...
+Base.:(-)(x::OwnedPtr{T}, y::Integer) where {T} = OwnedPtr{T}(pointer(x) - y, x.ctx)
+Base.:(+)(x::Integer, y::OwnedPtr) = y + x
+
 # simple conversions between Ptr and OwnedPtr are disallowed
 Base.convert{T}(::Type{Ptr{T}}, p::OwnedPtr{T}) = throw(InexactError())
 Base.convert{T}(::Type{OwnedPtr{T}}, p::Ptr{T}) = throw(InexactError())
 
 # unsafe ones are allowed
-Base.unsafe_convert{T}(::Type{Ptr{T}}, p::OwnedPtr{T}) = p.ptr
+Base.unsafe_convert{T1, T2}(::Type{Ptr{T1}}, p::OwnedPtr{T2}) = Ptr{T1}(p.ptr)
```
maleadt: Just drop `T2`, ie. `Base.unsafe_convert{T}(::Type{Ptr{T}}, p::OwnedPtr)`

src/memory.jl
```diff
@@ -38,6 +38,40 @@ function free(p::OwnedPtr)
...
+If reference count will get decreased to 0, `free_func(ptr, args...)` will get called
+"""
+function free_reference!(ptr::T) where T
+    _ptr = Base.unsafe_convert(Ptr{Void}, ptr)
+    if !haskey(_refcount_dictionary, _ptr)
+        error("Freeing unknown reference: $ptr")
+    end
+    refcount = _refcount_dictionary[_ptr]
+    if refcount == 0
+        error("Double free of reference: $ptr")
```
maleadt: if you make `free` handle the refcount, this case can't happen.

src/array.jl
maleadt: in addition, you can make `Mem.malloc` call `retain` such that you only need the additional `retain` call from the unsafe `CuArray` constructor.

maleadt (changes requested):

src/memory.jl
SimonDanisch: well, I still would like to have an error if it happens ;)

src/array.jl
SimonDanisch: I tried to separate these, because we're actually counting the Julia references to the same pointer, so I thought it might make sense to keep them in the constructor/finalizer... So Mem.free / Mem.alloc is considered to be more low level and could also be used without the creation of Julia objects + finalizer.

maleadt: >  could also be used without the creation of Julia objects + finalizer

Why wouldn't you be able to do that? Putting the refcounting logic in `memory.jl` would only imply that `free` will only be called once after doing a `malloc`, it wouldn't change the externally-facing API, ie. you don't need finalizers.

In addition, it would make it easy to re-use the refcounting logic, for when people want to define their own array types.

src/memory.jl
maleadt: maybe make it an `@assert` then, given that it's not a true error but a corrupt internal state

src/array.jl
SimonDanisch: okay fair enough...

maleadt:
@SimonDanisch Went with your proposal (isolated refcounting, not integrated with basic pointer management), but cleaned-up the API slightly. Thoughts?

SimonDanisch:
Looks good to me:)

maleadt (approved):

maleadt:
👍  Merging!

PR #55, opened 2017-08-24
Expand usage section in docs

una-dinosauria:
In the spirit of CUDArt's [README](https://github.com/JuliaGPU/CUDArt.jl/blob/master/README.md), which greatly helped me run my first CUDA kernels a couple of years ago.

Related: https://github.com/JuliaGPU/CUBLAS.jl/issues/34


maleadt:
Thanks for the contribution! Wouldn't it be better to put that content in the documentation instead, ie. [here](http://juliagpu.github.io/CUDAdrv.jl/latest/)? I purposefully stripped the README of any non-trivial information a while ago to avoid duplication between the README and the docs.

una-dinosauria:
Makes sense. I've updated the PR.

maleadt:
Great, thanks. I'll review up close next week (or just push some changes, if you don't mind). I'm not a big fan of the `list of modules` approach of the CUDArt docs...
Also, CI seems confused by this, and I think it will fail doctests (because you removed some trailing semicolons).

una-dinosauria:
`ci.maleadt.net` seems to be down

maleadt:
Does it? Works from here... What IP does it resolve to?

una-dinosauria:
`109.236.138.147`. I get an `ERR_TIMED_OUT` on the browser.

maleadt:
That is the correct IP indeed. Can you try `curl` on ports 80 and 8010? If not, I'll look into it tomorrow.
For now, here's the log of the build failure: https://pastebin.com/xzAUVJNn

MikeInnes:
There seems to be an error running the doctests – you can reproduce it locally with `Pkg.add("Documenter"); Pkg.test("CUDAdrv")`. Taking a quick look now so may be able to fix it.

MikeInnes:
Got it – thanks for the patch!

una-dinosauria:
Thanks @MikeInnes, I was kinda busy the last few days.
@maleadt, I was wondering if you could share what approach you like instead of modules?

maleadt:
> I was wondering if you could share what approach you like instead of modules?

Oh, I just meant the `init`/`mdinit`/`finit`/`mdclose` boilerplate. Looks like smelly code, but I haven't thought about a cleaner replacement.

Also, can you reach my CI now?

una-dinosauria:
Thanks. ci.maleadt.net is reachable for me now.

PR #56, opened 2017-08-31
Use Ref to denote GC-managed memory.

maleadt:


vchuravy (approved):

maleadt:
I'm not sure how to deal with functions accepting a `void*` though, eg. `cuMemAlloc`. I can allocate a `Ref{T}`, but how do I pass that as eg. a `Ref{Void}`? There doesn't seem to be a `convert(Ref{Void}, ::Ref)`, and `unsafe_convert(Ref{Void}, ::Ref` just yields a `Ptr{Void}` which probably doesn't offer any GC tracking after that.

This matters for eg. `cuMemcpyHtoD` where there is no further use of the argument after the `ccall`.

Maybe I should just pass a `Ref{T}`, but is that always compatible with functions accepting a `void*`?

yuyichao:
> I'm not sure how to deal with functions accepting a void* though, eg. cuMemAlloc

Leave them as is?

maleadt:
> Leave them as is?

Is that safe GC wise? Say I have:

```
function set{T}(var::CuGlobal{T}, val::T)
    val_ref = Ref{T}(val)
    ccall(:cuMemcpyHtoD, Cint, (Ptr{Void}, Ptr{Void}, Csize_t),
                               var.ptr, val_ref, var.nbytes)
end
```

Can't val_ref get collected before the ccall because it is passed as Ptr and not as Ref (https://docs.julialang.org/en/latest/manual/calling-c-and-fortran-code/#Some-Examples-of-C-Wrappers-1) ?

yuyichao:
`var.ptr` is wrong.

None of the changes in this PR is needed for GC safety.

maleadt:
> var.ptr is wrong.

Why would it be? It's a non-GC managed pointer allocated by `cuModuleGetGlobal`. I'm talking about passing `val`.

> None of the changes in this PR is needed for GC safety.

I know the bulk is unnecessary, but I added it to differentiate GC-managed from other pointers.
There's changes which, I think, have an affect on correctness too though, changing `Ptr{Cstring}` to `Ref{Cstring}`.

yuyichao:
In general, the only invalid thing to do is do pointer convertion yourself and by convertion I mean to get a pointer object explicitly from the julia object that manages it's memory. That means you must not call `pointer`, `pointer_from_objref`, `unsafe_convert` explicitly and also not loading a pointer field outside `unsafe_convert`.

> There's changes which, I think, have an affect on correctness too though, changing Ptr{Cstring} to Ref{Cstring}.

Nop.

yuyichao:
And in general, the type used for `ccall` argument does **NOT** affect the correctness for GC. They do have different semantics so it's not always correct to use one vs the other of course.

maleadt:
Seems like the nuances of when the GC is allowed to collect stuff is lost on me...

For example, in https://github.com/ihnorton/Clang.jl/issues/152 you mention changing:

```julia
function DAQmxLoadTask(taskName::Ptr{UInt8},taskHandle::Ptr{TaskHandle})
    ccall((:DAQmxLoadTask,NIDAQmx),int32,(Ptr{UInt8},Ptr{TaskHandle}),taskName,taskHandle)
end
```

to

```julia
function DAQmxLoadTask(taskName,taskHandle)
    ccall((:DAQmxLoadTask,NIDAQmx),int32,(Cstring,Ptr{TaskHandle}),taskName,taskHandle)
end
```

In this PR I similarly do (only talking about the `Cstring` change here):

```patch
     function (::Type{CuGlobal{T}}){T}(mod::CuModule, name::String)
         ptr_ref = Ref{Ptr{Void}}()
         nbytes_ref = Ref{Cssize_t}()
-        @apicall(:cuModuleGetGlobal, (Ptr{Ptr{Void}}, Ptr{Cssize_t}, CuModule_t, Ptr{Cchar}), 
+        @apicall(:cuModuleGetGlobal, (Ref{Ptr{Void}}, Ref{Cssize_t}, CuModule_t, Cstring),
                                      ptr_ref, nbytes_ref, mod, name)
```

Yet you mention `None of the changes in this PR is needed for GC safety.`.
What is the difference? (probably the signature?)


Also, in [the manual](https://docs.julialang.org/en/latest/manual/calling-c-and-fortran-code/#Some-Examples-of-C-Wrappers-1) there's an example mentioning passing an array that isn't used after the ccall with `Ref{...}` instead of `Ptr{...}`. I presumed that is because the GC memory can be collected after evaluating the arguments, but before the actual call happens. Is that correct?

Another example shows `Ref` in a similar fashion:

```julia
function permutation_free(p::Ref{gsl_permutation})
    ccall(
        (:gsl_permutation_free, :libgsl), # name of C function and library
        Void,                             # output type
        (Ref{gsl_permutation},),          # tuple of input types
        p                                 # name of Julia variable to pass in
    )
end
```

Given those 2 examples, I thought I would need to use `Ref` here as well:

```julia
function set{T}(var::CuGlobal{T}, val::T)
    val_ref = Ref{T}(val)
    ccall(:cuMemcpyHtoD, Cint, (..., Ptr{Void}, ...),
                               ..., val_ref, ...)
    # no further uses of val or val_ref
end
```

Why wouldn't I need the `Ref` here then?


Also, you mentioned `var.ptr is wrong` which I tried to "disprove" -- could you confirm whether my reasoning is correct?


Just trying to get a clear understanding of the GC, thanks for explaining.

maleadt:
Ah, didn't read  your additional reply...

yuyichao:
> In this PR I similarly do (only talking about the Cstring change here):

Well, you were talking about. `changing Ptr{Cstring} to Ref{Cstring}.`. In either case, it doesn't affect GC correctness. It may affect API correctness though and I've even mentioned that in the Clang.jl issue

> The Cstring vs Ptr{UInt8} could be a little subtle since the function doesn't always want a null-terminated string but using Cstring instead of Ptr{UInt8} should generally be safer.

maleadt:
Right, I understand the Cstring null-termination behaviour.
I just thought this change was required because the manual states `result_array must be declared to be of type Ref{Cdouble} and not Ptr{Cdouble}` (where `declared` is talking about the `ccall` signature because `result_array` itself is a `Vector{Cdouble}`) vs. you saying `the type used for ccall argument does NOT affect the correctness for GC`...

yuyichao:
If you do not have custom cconvert/unsafe_convert defs, the only technical advantage of using `Ref` on `ccall` is that it can take an object and pass it by ref (for fortran basically), for everything else in base that works with a `Ref` ccall type, `Ptr` should work.

After the `unsafe_convert`/`cconvert` lowering, `Ref` and `Ptr` are treated exactly the same which basically mean that other than the available `unsafe_convert` and `cconvert` method, there are no differenct at all between the two. `Ref` is recommended in the doc mainly because it is somehow less likely to make people do the explicit convertion (AFAICT) and is slightly more general in most cases with normal types. This is not always true of course...

yuyichao:
> result_array must be declared to be of type Ref{Cdouble} and not Ptr{Cdouble}

That's wrong.... The recommendation isn't wrong though so I don't feel like fixing it. (Plus, I'm not sure how to write that doc more clearly...)

maleadt:
OK thanks.

Out of curiosity, given:

```julia
function set{T}(var::CuGlobal{T}, val::T)
    val_ref = Ref{T}(val)
    ccall(:cuMemcpyHtoD, Cint, (..., Ptr{Void}, ...),
                               ..., val_ref, ...)
    # no further uses of val or val_ref
end
```

What keeps `val_ref` from getting collected before the pointer resulting from `unsafe_convert(cconvert(...))` is actually passed to `cuMemcpyHtoD`, given that there's no uses (except for the `ccall`, but does the GC know about that?)

yuyichao:
`cconvert` should do the right thing.

maleadt:
> cconvert should do the right thing.

... what does that mean? What is `the right thing`? Following that example, `cconvert(Ptr{Void}, ::Ref{Int})` does nothing and just defers the conversion to `Ptr{Void}` to `unsafe_convert`. So where does it get rooted for the `ccall`?

yuyichao:
```
help?> Base.cconvert
  cconvert(T,x)

  Convert x to a value to be passed to C code as type T, typically by calling convert(T, x).

  In cases where x cannot be safely converted to T, unlike convert, cconvert may return an object
  of a type different from T, which however is suitable for unsafe_convert to handle. The result
  of this function should be kept valid (for the GC) until the result of unsafe_convert is not
  needed anymore. This can be used to allocate memory that will be accessed by the ccall. If
  multiple objects need to be allocated, a tuple of the objects can be used as return value.

  Neither convert nor cconvert should take a Julia object and turn it into a Ptr.
```

It returns the object(s) that needs to be rooted.

maleadt:
Ah, so the semantics of `ccall` are that it roots the result of `cconvert` on all of its arguments? That is useful info, if not completely new to me (and IMO not clear from the relevant docstrings).

PR #60, opened 2017-09-18
Replace OwnedPtr with untyped/unsized Buffer

maleadt:


maleadt:
@SimonDanisch @MikeInnes this PR is relevant to you, because it gets rid of `OwnedPtr` and replaces it with a simpler `Buffer` type. A `Buffer` is meant to be opaque, and doesn't support much (eg. I removed pointer arithmetic).

Also relevant, I'm thinking of removing `pointer(::CuArray)` because it would bypass the underlying `Buffer` (which keeps the context alive). But everything still works because of how `cudacall` converts arguments:

```julia
julia> a = CuArray([1])
1-element Array{Int64,1}:
 1

julia> T = Ptr{Int}
Ptr{Int64}

julia> Base.unsafe_convert(T, Base.cconvert(T, a))
Ptr{Int64} @0x0000000b05400000
```

maleadt:
This is of course a step-up to getting rid of `CuArray` in its entirety, and putting some custom GC-like layer at the allocation sites.

MikeInnes:
:+1:

I think getting rid of `pointer` is the right move; I can't imagine where you'd need it in generic code, and the ccall code that expects a CPU pointer should just error out rather than segfaulting.

maleadt:
Had another go at this. Removed the object-based API and made it more strict (only accepting arrays) because otherwise `alloc(1)` would be ambiguous (allocate for object `1` or allocate 1 byte), etc.
Array-based API seems more clear too, and removes the issue with immutable objects we can't `Ref`.

@MikeInnes @vchuravy Care to have a(nother) look? This is going to be the one and only memory API in CUDAdrv, as I plan on removing CuArray and redirecting to CuArrays.jl, so it better be good 😄

maleadt:
https://devblogs.nvidia.com/parallelforall/gpu-computing-julia-programming-language/#comment-3629700604

It's maybe a good time to think about adding asynchronous/stream-bound transfers. Figuring out the actual programming interface is a problem to be solved in CuArrays.jl then.

maleadt:
OK, this is getting in pretty good shape. Will be testing with CUDAnative and committing later today.

MikeInnes:
Sorry I didn't get to providing feedback yet – I'll take a look next week, although I'm sure it's solid as is.

RE async, I think we can lean towards being async by default, everywhere that's feasible. I can see why the C syncs in some cases – given the difficulty of freeing CPU pointers at the right time – but we can handle that pretty easily with the GC, I think.

maleadt:
Downloads being async seems counter-intuitive for REPL usage, eg. `Mem.download(Int, buf)` would yield an vector that doesn't necessarily contain relevant data straight-away. So I'd prefer to keep it synchronizing by default, if only to mimic the C API.

maleadt:
@MikeInnes did you have time to look at the PR? Just because I'd like to tag CUDAdrv, and you'll probably need to adapt CuArrays (at some point).

una-dinosauria:
Hi! I have some code that was using `OwnedPtr` and now crashes. What's the new way to handle this? I'm still using Julia 0.6 and I'd prefer to not upgrade to 0.7 in the meantime, but if that's the only way then oh well.

maleadt:
`Buffer` is, and this is also in the 0.6 version. See the tests or examples folder; feel free to ask on discourse on Slack if you have other issues.

PR #62, opened 2017-09-20
Fix referencing of kernel arguments.

maleadt:
Bitstypes weren't compatible with pointer_from_objref,
and the much more aggressive GC on 0.7 made it fail.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -94,7 +94,7 @@ end
     push!(ex.args, :(kernelParams = [$(arg_ptrs...)]))
 
     # root the argument boxes to the array of pointers,
-    # because the GC doesn't know both are linked
+    # keeping them alive across the call to `cuLaunchKernel`
     if VERSION >= v"0.7.0-DEV.1850"
         push!(ex.args, :(Base.@gc_preserve $(arg_refs...) kernelParams))
     end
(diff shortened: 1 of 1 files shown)


maleadt:
Requires JuliaLang/julia#23783

maleadt:
Also ref JuliaLang/julia#23744

yuyichao:
> much more aggressive GC

Much more aggressive CG (codegen), not GC.

maleadt:
> Much more aggressive CG (codegen), not GC.

Hah, yeah ok that is a better description indeed.

Also, could you have a look at the [TODO](https://github.com/JuliaGPU/CUDAdrv.jl/blob/94856282c318a6528e686ee9b922775a09c222f6/src/execution.jl#L202-L203) I've put in the code? So essentially I have my own replacement for `ccall`, `cudacall`, which generates code much like `lower-ccall`:

```
gensym1 = Base.cconvert(typ1, arg1)
...
launch(..., Base.unsafe_convert(typ1, gensym1), ...)
```

I would think I also need to `@gc_preserve` the result of `unsafe_convert` to `gensym1`, but codegen doesn't seem to optimize this away.

yuyichao:
https://github.com/JuliaLang/julia/pull/23783

maleadt:
No, I meant, do I _need_ to use `@gc_preserve` for that use of `unsafe_convert`, because codegen currently doesn't seem to optimize the "parent" away (something it does in other cases as I reported in https://github.com/JuliaLang/julia/pull/23744#issuecomment-330536068 hence I knew I needed `@gc_preserve`).
Also, the lowered code for a similar `ccall` doesn't contain any explicit rooting, but maybe that uses another mechanism to root the result of `cconvert`?

yuyichao:
You have to use `@gc_preserve` whenever you use `unsafe_convert` (or equivalently the gc root list in `ccall`)

> codegen currently doesn't seem to optimize the "parent" away

That's exactly what the PR I linked does.

yuyichao:
i.e. https://github.com/JuliaLang/julia/pull/23783/files#diff-d14f5feef27442143be917f90c9b9079

yuyichao:
Though you did make me realize that I forgot to actually test for the allocation ellision...

PR #77, opened 2018-01-04
Allow transfer of array views backed by contiguous memory (fix #76)

### LINKED ISSUE #76: copy! methods for views with linear indices
I have an application in which an (expensive) linear operator is implemented on the GPU. Methods such as `GMRES` from `IterativeSolvers` attempt to apply the operator to a RHS which is a view of another larger array. Since there are no `copy!` methods for views, a naive implementation fails.

In the case of GMRES (and one would imagine most performance critical applications), the views will be an appropriate sized column of a matrix, thus the memory is linear and the copy operation could presumably be performed without too much pain.

I am not that familiar with the `SubArray` architecture - would it be to implement a `copy!` method which accepts a view, so long as the indices are linear and the dimensions correct?

samuelpowell:
Ref #76

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -174,7 +174,7 @@ const copyfun = VERSION >= v"0.7.0-DEV.3057" ? :(copyto!) : :(copy!)
     Copy an array view from a host array `src` to a device array `dst` in place. Both arrays
     should have an equal length, and the stide of the view must be unity in all dimensions.
     """
-    function Base.$copyfun(dst::CuArray{T}, src::SubArray{T,N,A,I,true}) where T
+    function Base.$copyfun(dst::CuArray{T}, src::SubArray{T,N,A,I,true}) where {T,N,A,I}
         if length(dst) != length(src)
             throw(ArgumentError("Inconsistent array length."))
         end
@@ -191,7 +191,7 @@ const copyfun = VERSION >= v"0.7.0-DEV.3057" ? :(copyto!) : :(copy!)
     Copy an array from a device array `src` to a host array view `dst` in place. Both arrays
     should have an equal length, and the stride of the view must be unity in all dimensions.
     """
-    function Base.$copyfun(dst::SubArray{T,N,A,I,true}, src::CuArray{T}) where T
+    function Base.$copyfun(dst::SubArray{T,N,A,I,true}, src::CuArray{T}) where {T,N,A,I}
         if length(dst) != length(src)
             throw(ArgumentError("Inconsistent array length."))
         end
--- a/src/array.jl
+++ b/src/array.jl
@@ -169,37 +169,45 @@ const copyfun = VERSION >= v"0.7.0-DEV.3057" ? :(copyto!) : :(copy!)
     end
 
     """
-        $copyfun{T}(dst::CuArray{T}, src::SubArray{T,N,A,I,true})
+        $copyfun{T}(dst::CuArray{T}, src::SubArray{T,N,<:DenseArray,I,true})
 
     Copy an array view from a host array `src` to a device array `dst` in place. Both arrays
-    should have an equal length, and the stide of the view must be unity in all dimensions.
+    should have an equal length, and the view must have a contiguous memory layout.
     """
-    function Base.$copyfun(dst::CuArray{T}, src::SubArray{T,N,A,I,true}) where {T,N,A,I}
+    function Base.$copyfun(dst::CuArray{T}, src::SubArray{T,N,<:DenseArray,I,true}) where {T,N,I}
         if length(dst) != length(src)
             throw(ArgumentError("Inconsistent array length."))
         end
-        if any(strides(src) .!= 1)
-            throw(ArgumentError("Transfers to an array view require unit stride."))
+        if any(strides(src) .!= strides(parent(src)))
+            throw(ArgumentError("Transfers from an array view require contiguous memory layout."))
         end
         Mem.upload!(dst.buf, pointer(src), length(src) * sizeof(T))
         return dst
+    end
+
+    function Base.$copyfun(dst::CuArray, src::SubArray)
+        throw(ArgumentError("Transfers from an array view require a contiguous memory layout."))
     end
 
     """
         $copyfun{T}(dst::SubArray{T,N,A,I,true}, src::CuArray{T})
 
     Copy an array from a device array `src` to a host array view `dst` in place. Both arrays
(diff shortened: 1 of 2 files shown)


samuelpowell:
This approach seems to work okay, but the use case may be a bit niche. 

@maleadt if you have an appetite to merge this I will add tests.

src/array.jl
```diff
@@ -167,6 +167,41 @@ const copyfun = VERSION >= v"0.7.0-DEV.3057" ? :(copyto!) : :(copy!)
...
+        $copyfun{T}(dst::CuArray{T}, src::SubArray{T,N,A,I,true})
+
+    Copy an array view from a host array `src` to a device array `dst` in place. Both arrays
+    should have an equal length, and the stide of the view must be unity in all dimensions.
+    """
+    function Base.$copyfun(dst::CuArray{T}, src::SubArray{T,N,A,I,true}) where {T,N,A,I}
+        if length(dst) != length(src)
+            throw(ArgumentError("Inconsistent array length."))
+        end
+        if any(strides(src) .!= 1)
```
timholy: This will fail for any multidimensional array, where even a dense array has a stride bigger than one along higher dimensions.

`strides` is not entirely safe yet, but see https://github.com/JuliaLang/julia/pull/25321 (which should make it safe). If you rely on an unsafe `strides`, you might get wrong answers or segfaults.

samuelpowell: @timholy good point, thanks. I was concentrating on the particular problem I have in #76 and hadn't properly considered this. 

I could fix this to work for views which only restrict the trailing dimension to a contiguous region, but that's too peculiar. Alternatively, I could restrict the method only to operate on views which result in a vector, which would still be useful. 

That said, if `strides` isn't safe, then that might be a showstopper. Feel free to close the PR if that's the case - I can always implement these methods privately in my package until things stabilise.

maleadt:
> @maleadt if you have an appetite to merge this I will add tests.

No objections per se, but since we're transitioning away from CUDAdrv.CuArrays (just waiting for 0.7 to stabilize and be supported by GPUArrays) it might be better to add this to CuArrays.jl.
But if you'd like this functionality to be in CUDAdrv for a couple of releases still, feel free to push this forward 🙂

samuelpowell:
I haven't yet started the transition to 0.7, so it would be useful to me if this were available here whilst I am still working on 0.6/updating packages. I suspect a more generic/well planned interface which takes account of various array types would be better for the future, and have seen a few related issues and PRs at CuArrays.jl.

For the moment, I have updated the PR to attempt to ensure that it only attempts to copy views which represent a contiguous memory backed region, and added some tests accordingly. @timholy does this address your concerns, or is it still problematic?

timholy:
Yes, this seems safe.

samuelpowell:
Thanks @timholy

maleadt:
Thanks. CI failure unrelated, probably because this PR isn't based on master.

rveltz:
Hi, 

I am having a similar problem. Is it really solved?

```julia

julia> phi0 , flag = @time IterativeSolvers.gmres(Jl,v,tol = 1e-6,log = true,verbose=true,restart=200)
ERROR: conversion to pointer not defined for CuArray{Float32,1}
Stacktrace:
 [1] copy!(::SubArray{Float32,1,Array{Float32,2},Tuple{Base.Slice{Base.OneTo{Int64}},Int64},true}, ::CuArray{Float32,1}) at /user/rveltz/home/.julia/v0.6/GPUArrays/src/abstractarray.jl:120
 [2] #init!#55(::Bool, ::Function, ::IterativeSolvers.ArnoldiDecomp{Float32,LinearMaps.FunctionMap{Float32,JacobianNF,Void}}, ::Array{Float32,1}, ::CuArray{Float32,1}, ::IterativeSolvers.Identity, ::Array{Float32,1}) at /user/rveltz/home/.julia/v0.6/IterativeSolvers/src/gmres.jl:219
 [3] (::IterativeSolvers.#kw##init!)(::Array{Any,1}, ::IterativeSolvers.#init!, ::IterativeSolvers.ArnoldiDecomp{Float32,LinearMaps.FunctionMap{Float32,JacobianNF,Void}}, ::Array{Float32,1}, ::CuArray{Float32,1}, ::IterativeSolvers.Identity, ::Array{Float32,1}) at ./<missing>:0
 [4] #gmres_iterable!#52(::IterativeSolvers.Identity, ::IterativeSolvers.Identity, ::Float64, ::Int64, ::Int64, ::Bool, ::Function, ::Array{Float32,1}, ::LinearMaps.FunctionMap{Float32,JacobianNF,Void}, ::CuArray{Float32,1}) at /user/rveltz/home/.julia/v0.6/IterativeSolvers/src/gmres.jl:118
 [5] (::IterativeSolvers.#kw##gmres_iterable!)(::Array{Any,1}, ::IterativeSolvers.#gmres_iterable!, ::Array{Float32,1}, ::LinearMaps.FunctionMap{Float32,JacobianNF,Void}, ::CuArray{Float32,1}) at ./<missing>:0
 [6] #gmres!#54(::IterativeSolvers.Identity, ::IterativeSolvers.Identity, ::Float64, ::Int64, ::Int64, ::Bool, ::Bool, ::Bool, ::IterativeSolvers.#gmres!, ::Array{Float32,1}, ::LinearMaps.FunctionMap{Float32,JacobianNF,Void}, ::CuArray{Float32,1}) at /user/rveltz/home/.julia/v0.6/IterativeSolvers/src/gmres.jl:185
 [7] (::IterativeSolvers.#kw##gmres!)(::Array{Any,1}, ::IterativeSolvers.#gmres!, ::Array{Float32,1}, ::LinearMaps.FunctionMap{Float32,JacobianNF,Void}, ::CuArray{Float32,1}) at ./<missing>:0
 [8] #gmres#53(::Array{Any,1}, ::Function, ::LinearMaps.FunctionMap{Float32,JacobianNF,Void}, ::CuArray{Float32,1}) at /user/rveltz/home/.julia/v0.6/IterativeSolvers/src/gmres.jl:134
 [9] (::IterativeSolvers.#kw##gmres)(::Array{Any,1}, ::IterativeSolvers.#gmres, ::LinearMaps.FunctionMap{Float32,JacobianNF,Void}, ::CuArray{Float32,1}) at ./<missing>:0
```

samuelpowell:
It was certainly working for me at the time, though I haven't used it recently. I suggest opening an issue with a minimum working example and xref accordingly.

rveltz:
Thank you for your answer. I will do as suggested.

Best regards,

Romain

PR #81, opened 2018-01-15
Fix Compat requirement.

### LINKED ISSUE #80: Cvoid not defined when importing package
I am working on Ubuntu 17.10, CUDA 9.1, and Julia 0.6.2. My CUDA drivers work: I just run a C++ code with CUDA.

My process has been:

```Pkg.add("CUDAdrv")```

works fine, when I try again 

```
INFO: Package CUDAdrv is already installed
```

```Pkg. build("CUDAdrv")```

with the result:

```
INFO: Building CUDAdrv
WARNING: Found multiple CUDA driver installations: /usr/lib/x86_64-linux_gnu and /usr
```

Next, I type

```
using CUDAdrv
INFO: Precompiling module CUDAdrv
ERROR: LoadError: LoadError:

UndefVarError:Cvoid not defined
Stacktrace:
```
 ------bunch of additional stuff (I am not writing this in the same computer, so it is hard to copy the whole thing-------
```
failed to precompile CUDAdrv to <my directory here>
```

Same outcome if I skip `Pkg. build("CUDAdrv")` and invoke directly using CUDAdrv.

I saw this https://github.com/JuliaGPU/CUDAdrv.jl/issues/68 but it does not seem to be the same problem (even if the error message is the same with `Pkg. build("CUDAdrv")`).

Any suggestions?

I'd like to use the package for some very basic teaching of Julia with GPUs

maleadt:
0.46.0 is necessary for the Void deprecation (fixes #80).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.6
-Compat 0.39.0
+Compat 0.46.0
 CUDAapi 0.3.0


maleadt:
CI failure unrelated.

PR #93, opened 2018-02-02
expose UVM API

vchuravy:
I mocked this up to test how UVM would impact an application.
This is low priority for me so  I won't be spending much time on it.

### DIFF
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -5,7 +5,7 @@ export Mem
 module Mem
 
 using CUDAdrv
-import CUDAdrv: @apicall, CuStream_t
+import CUDAdrv: @apicall, CuStream_t, CuDevice_t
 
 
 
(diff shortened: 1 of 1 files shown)


maleadt:
I think we need a Buffer type in Base (eg. JuliaLang/julia#12447) before tackling this.

vchuravy:
Yeah it is much less useful if you only use it for GPU memory

maleadt:
Maybe we should make this part of the type, so that CuArray can have an optimized convert to Array that just does unsafe_wrap.
EDIT: or just make CuArray use this all the time.

maleadt:
Rebased.

maleadt:
> Yeah it is much less useful if you only use it for GPU memory

So that turned out to be false, and unified memory is very useful for handling multi-GPU setups.
Let's merge if CI agrees.

PR #98, opened 2018-03-16
fix depwarns for 07

SimonDanisch:


src/array.jl
```diff
@@ -10,7 +10,7 @@ export
     CuArray{T}(dims)
     CuArray{T,N}(dims)
 
-Construct an uninitialized `N`-dimensional dense CUDA array with element type `T`, where `N`
+Construct an undef `N`-dimensional dense CUDA array with element type `T`, where `N`
```
SimonDanisch: do we keep this change since it corresponses to Julia.undef?

maleadt: No, Base doesn't document it like that

```
  Array{T}(undef, dims)
  Array{T,N}(undef, dims)

  Construct an uninitialized N-dimensional Array containing elements of type T
```

Issue #122, opened 2019-01-16
Support for cudaHostAlloc() cudaHostGetDevicePtr() cudaFreeHost() including flag optionss


wsphillips:
In reference to this [post](https://discourse.julialang.org/t/efficiency-when-handling-jobs-larger-than-vram/19570/3)

VERY VERY much want/would appreciate/be ever thankful for [cudaHostAlloc() + flags, cudaHostGetDevicePtr() and cudFreeHost()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html) to improve HtD and DtH transfer rates. It looks like you have some in your api dictionary already. If you provide wrappers that return Buffer objects I can have a stab at making some support functions for a PR and document it for posterity. I am thinking handling functions could be implemented safely using just Base  `pointer_from_objref` and `unsafe_copyto!` 

I would also need a simple dispatch for `Mem.upload!` and `Mem.download!` that allows for two buffer objects (in this case, one on host, one on device) with async and stream settings as usual.

[Flags for HostAlloc](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gb65da58f444e7230d3322b6126bb4902) = default, portable, mapped and writecombine

PR #123, opened 2019-01-20
Implementation of pinned memory allocation

wsphillips:
Re: #122

This seems to work. Only problem: I have to specify `libc.so.6` in order to call standard C-library functions. Allowing Julia to try and find it seems to fail. Any advice on how to make it "run for anyone" appreciated.
Another problem may be checking for type compatibility when transferring data to/from the buffer. At present this all assumes you know input and output types.

Have a look. Please do revise as necessary! If you think its okay, I can add some documentation for posterity and develop a simple benchmark to look at.

### DIFF
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -161,7 +161,6 @@ function freehost(buf::Buffer)
     end
     return
 end
-
 """
     alloc(bytes::Integer)
 
@@ -255,12 +254,41 @@ function upload!(dst::Buffer, src::Ref, nbytes::Integer,
     end
 end
 
+#When using pinned host memory, src can be a buffer
+function upload!(dst::Buffer, src::Buffer, nbytes::Integer,
+                 stream::CuStream=CuDefaultStream(); async::Bool=false)
+    if async
+        @apicall(:cuMemcpyHtoDAsync,
+                 (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t, CuStream_t),
+                 dst, src, nbytes, stream)
+    else
+        @assert stream==CuDefaultStream()
+        @apicall(:cuMemcpyHtoD,
+                 (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t),
+                 dst, src, nbytes)
+    end
+end
+
 """
     download!(dst::Ref, src::Buffer, nbytes::Integer, [stream=CuDefaultStream()]; async=false)
 
 Download `nbytes` memory from `src` on the device to `src` on the host.
 """
 function download!(dst::Ref, src::Buffer, nbytes::Integer,
+                   stream::CuStream=CuDefaultStream(); async::Bool=false)
+    if async
+        @apicall(:cuMemcpyDtoHAsync,
+                 (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t, CuStream_t),
+                 dst, src, nbytes, stream)
+    else
+        @assert stream==CuDefaultStream()
+        @apicall(:cuMemcpyDtoH,
+                 (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t),
+                 dst, src, nbytes)
+    end
+end
+# When using host pinned memory, destination can be a buffer
+function download!(dst::Buffer, src::Buffer, nbytes::Integer,
                    stream::CuStream=CuDefaultStream(); async::Bool=false)
     if async
         @apicall(:cuMemcpyDtoHAsync,
@@ -340,7 +368,21 @@ function download!(dst::AbstractArray, src::Buffer,
     return
 end
 
+function loadPinned(dst::Buffer, src::AbstractArray)
+    if (sizeof(src) > dst.bytesize) || (sizeof(src) < dst.bytesize)
+        throw(ArgumentError("size of destination does not match size of source (bytes)"))
+    end
+    srcptr = Ref(src, 1)
+    ccall((:memcpy, "libc.so.6"),Ptr{Cvoid},(Ptr{Cvoid},Ptr{Cvoid}, Csize_t), dst.ptr, srcptr, dst.bytesize)
+    end
 
+function unloadPinned(dst::AbstractArray, src::Buffer)
+    if (sizeof(dst) > src.bytesize) || (sizeof(dst) < src.bytesize)
+        throw(ArgumentError("size of destination does not match size of source (bytes)"))
+    end
+    dstptr = Ref(dst,1)
+    ccall((:memcpy, "libc.so.6"),Ptr{Cvoid},(Ptr{Cvoid},Ptr{Cvoid}, Csize_t), dstptr, src.ptr, src.bytesize)
+end
 ## type based
 
 function check_type(::Type{Buffer}, T)
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -144,6 +144,23 @@ function transfer end
 @enum(CUmem_attach, ATTACH_GLOBAL = 0x01,
                     ATTACH_HOST   = 0x02)
                     #ATTACH_SINGLE = 0x04) # Defined but not valid
+@enum(CUmem_hostalloc, default       = 0x00,
+                       mapped        = 0x02,
+                       portable      = 0x01,
+                       writecombined = 0x04)
+
+function hostalloc(bytesize::Integer, flags::CUmem_hostalloc=default)
+    ptr_ref = Ref{Ptr{Cvoid}}()
+    @apicall(:cuMemAllocHost, (Ptr{Ptr{Cvoid}}, Csize_t, Cuint), ptr_ref, bytesize, flags)
+    return Buffer(ptr_ref[], bytesize, CuCurrentContext())
+end
+
+function freehost(buf::Buffer)
+    if buf.ptr != C_NULL
+    @apicall(:cuMemFreeHost, (Ptr{Cvoid},), buf.ptr)
+    end
+    return
+end
 
 """
     alloc(bytes::Integer)
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -25,8 +25,6 @@ function view(buf::Buffer, bytes::Int)
     return Mem.Buffer(buf.ptr+bytes, buf.bytesize-bytes, buf.ctx)
 end
 
-
-
 ## refcounting
 
 const refcounts = Dict{Buffer, Int}()
@@ -59,7 +57,6 @@ function release(buf::Buffer)
     refcounts[buf] = refcount - 1
     return refcount==1
 end
-
 
 ## memory info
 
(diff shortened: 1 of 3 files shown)


src/memory.jl
```diff
@@ -323,7 +368,21 @@ function download!(dst::AbstractArray, src::Buffer,
     return
 end
 
+function loadPinned(dst::Buffer, src::AbstractArray)
+    if (sizeof(src) > dst.bytesize) || (sizeof(src) < dst.bytesize)
+        throw(ArgumentError("size of destination does not match size of source (bytes)"))
+    end
+    srcptr = Ref(src, 1)
+    ccall((:memcpy, "libc.so.6"),Ptr{Cvoid},(Ptr{Cvoid},Ptr{Cvoid}, Csize_t), dst.ptr, srcptr, dst.bytesize)
```
vchuravy:
```suggestion
    ccall(:memcpy, Ptr{Cvoid}, (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t), dst.ptr, srcptr, dst.bytesize)
```

src/memory.jl
```diff
@@ -323,7 +368,21 @@ function download!(dst::AbstractArray, src::Buffer,
...
+    end
 
+function unloadPinned(dst::AbstractArray, src::Buffer)
+    if (sizeof(dst) > src.bytesize) || (sizeof(dst) < src.bytesize)
+        throw(ArgumentError("size of destination does not match size of source (bytes)"))
+    end
+    dstptr = Ref(dst,1)
+    ccall((:memcpy, "libc.so.6"),Ptr{Cvoid},(Ptr{Cvoid},Ptr{Cvoid}, Csize_t), dstptr, src.ptr, src.bytesize)
```
vchuravy:
```suggestion
    ccall(:memcpy, Ptr{Cvoid}, (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t), dstptr, src.ptr, src.bytesize)
```

src/memory.jl
```diff
@@ -256,6 +287,20 @@ function download!(dst::Ref, src::Buffer, nbytes::Integer,
                  dst, src, nbytes)
     end
 end
+# When using host pinned memory, destination can be a buffer
```
vchuravy:
```suggestion

# When using host pinned memory, destination can be a buffer
```

src/memory.jl
```diff
@@ -144,7 +144,23 @@ function transfer end
 @enum(CUmem_attach, ATTACH_GLOBAL = 0x01,
                     ATTACH_HOST   = 0x02)
                     #ATTACH_SINGLE = 0x04) # Defined but not valid
+@enum(CUmem_hostalloc, default       = 0x00,
```
vchuravy, line 145:
```suggestion

@enum(CUmem_hostalloc, default       = 0x00,
```

vchuravy (commented):
I am not sure I like reusing `Buffer` this way.

Last I had to do this, I just used `unsafe_wrap(Array, convert(Ptr{T}, r_ptr[]), dims, #=own=#false)`, but I am not sure that is a better API

--edit--

also tests would be good!

wsphillips:
I am working on some testing and have put together a way of wrapping buffer objects with CuArrays that simplifies moving data around and allows slicing with views.

Short story is using pinned memory + streams + async transfer you can get overlap between kernel execution, upload and download simultaneously that wasn't quite possible before. Right now I have everything in a separate module that just extends CUDAdrv/CuArrays so you can test it without having to merge in any code. Will post soon.

wsphillips:
Hey! 
Check out the zip file attached for some demos of the proposed enhancements.

Put these scripts into your working dir in Julia and just run them from Main. This should extend CUDAdrv and CuArrays without needing to make a separate build. Step through ```ffttest.jl``` block by block to see additions in action. Let me know what you guys think. I incorporated a "non-buffer" means of handling pinned memory as suggested by @vchuravy Obvious needed fix is a way of distinguishing CuArrays containing pinned memory from CuArrays that are GPU memory.

```bifft.jl``` is example usage of the extended functions 
```cufftext.jl``` contains just a couple API extensions for CuArrays.CUFFT 
```host.jl``` contains all the functions I have found useful for working with pinned memory and concurrency.


[pinnedmem.zip](https://github.com/JuliaGPU/CUDAdrv.jl/files/2795917/pinnedmem.zip)

Cheers!

**Comment by wsphillips** [comment]
Also summary: I can get 40-50% bandwidth increase using pinned memory transfers. 
Also aids concurrency. With the functions included I can get huge stacks of data to crunch with nice overlap like this:
screenshot from 2019-01-25 12-55-36

Greatly improves efficiency when cycling data onto the GPU that is much bigger than can be held at once.

File: src/memory.jl [hunk 5]
```diff
@@ -145,6 +142,106 @@ function transfer end
                     ATTACH_HOST   = 0x02)
                     #ATTACH_SINGLE = 0x04) # Defined but not valid
 
+@enum(CUmem_hostalloc, default       = 0x00,
```
wsphillips, line 145: So I know this is a little weird using CuArrays within CUDAdrv. Not sure how else to go about this, but it does make the handling less awkward than using buffer objects directly.

Here its mainly so you can see how it's mirroring the functionality. Ultimately if its using CuArrays to do so, it might be better off as a separate module in CuArrays? 

I am thinking CuArray objects could contain a boolean that flags whether its host allocated memory? (Or something like that...for type stability when uploading and downloading)

wsphillips:
> I am not sure I like reusing `Buffer` this way.
> 
> Last I had to do this, I just used `unsafe_wrap(Array, convert(Ptr{T}, r_ptr[]), dims, #=own=#false)`, but I am not sure that is a better API
> 
See what I have done with CuArrays as wrappers. Its a "mock" solution on my fork right now...not truly implemented. (i.e. these edits won't work in place atm for obvious reasons) Instead, see the extensions in the test files I have added for a working/testable variant that sits outside the source packages.

src/memory.jl
```diff
@@ -145,6 +142,106 @@ function transfer end
...
+                       portable      = 0x01,
+                       writecombined = 0x04)
+
+# Currently the pointer gets wrapped by a CuArray, but at present there's no 
+# implementation of a method for distinguishing regular GPU CuArrays from 
+# CPU-pinned memory CuArrays. 
+""" hotalloc() 
+allocates page-locked/pinned memory registered by the GPU for higher 
+bandwidth transfer and better compatibility with async methods. """
+function hostalloc(src::AbstractArray{T}; flags::CUmem_hostalloc=default) where T
```
vchuravy, line 156: What is `src` used for? If you just need `T` you can use `::Type{T}` and the dims independently

src/memory.jl
vchuravy: So I think the main question for me is.

```
A = hostalloc(Float32, 1024, 32)
B = CuArray{Float32}(undef, 1024, 32)
```

Given that what happens in `A .+ A` and `A .+ B`. Since they are both a CuArray this will dispatch to GPU code.

src/memory.jl
```diff
@@ -145,6 +142,106 @@ function transfer end
...
+function pin(dst::CuArray, src::AbstractArray)
+    if eltype(src) !== eltype(dst)
+        throw(ArgumentError("type of dst does not match type of src"))
+    end
+    if (sizeof(src) > sizeof(dst)) || (sizeof(src) < sizeof(dst))
+        throw(ArgumentError("size of destination does not match size source (bytes)"))
+    end
+    srcptr = Ref(src,1)
+    dstbuf = CuArrays.buffer(dst)
+    ccall(:memcpy, Nothing, (Ptr{Cvoid}, Ptr{Cvoid}, Csize_t), dstbuf.ptr,
```
vchuravy, line 186: This will fail if it is GPU memory right?

vchuravy:
Can you put the examples up as a [Gist](https://gist.github.com/). Thanks

wsphillips:
> 
> 
> Can you put the examples up as a [Gist](https://gist.github.com/). Thanks

https://gist.github.com/wsphillips/376b4dc65a9e4982252e91af6c41c32f

src/memory.jl
wsphillips: Sure. I will have a look. Admittedly I'm fairly new to Julia so there's likely quite a few inefficient pieces--partly due to naivete and partly wanting to just get it working for proof of principle. I will do my best to fix whatever you point out. Suggestions like these are super helpful!

src/memory.jl
wsphillips: Yes. I didn't want to start messing with the CuArrays type definition without getting some input. 
Wrapping the paged-locked memory as an unowned CuArray makes it easier to handle/view, pass to api calls, permits views compatibility and overall seems much more stable. However it opens it up to aberrant dispatch. 

I actually don't know that using GPU code will cause it to seriously fail though. I haven't tried intentionally doing unintended actions like that yet...

src/memory.jl
wsphillips: Not sure. Likely yes. I would like to find a way to prevent mixup of GPU/pinned memory. Again, some sort of tweak to the CuArrays type definition that gets left unused until it gets passed into a function designed for pinned memory might do the trick.

vchuravy: Right the way to prevent that mix-up would be to not use `CuArray`, but to use `Array` (or add `CuHostArray`, but that is a bunch of extra work).

wsphillips:
> > Can you put the examples up as a [Gist](https://gist.github.com/). Thanks
> 
> https://gist.github.com/wsphillips/376b4dc65a9e4982252e91af6c41c32f

Gist updated after @maleadt helped me sort out memory/plan handle issues. Demo of pinned memory still relevant for PR review/revision.

maleadt:
Sorry for taking so long to review this; I've been focusing on some more pressing matters. I'll try and have a look next week, but this PR needs to be updated for the new CUDAdrv pointer types anyway (I can do that when reviewing).

wsphillips:
Sure. No worries about the delay--as said I have a temporary fix via the modules I built in the Gist. 

I guess the major questions are:

1. How should pinned memory be wrapped? Array-like object or buffer object?
2. If array object, how does it remain type safe?
3. Is it more suited as an extension in CuArrays than CUDAdrv?

Thanks for taking the time to have a look!

wsphillips:
@maleadt I can see you have been hard at work with the new updates! I've been busy myself with grant writing but now I should have some time again to revisit/rethink this.

Do you think the new CudaPtr might be useful here?

maleadt:
Pinned memory should be fine and just use `Ptr` -- unified memory is a different deal (see https://github.com/JuliaGPU/CUDAdrv.jl/pull/125#issuecomment-460621655). Now that the dust has settled on these low-level changes, I'll have a look next week at your PRs!

maleadt:
Finally some time to give this a proper look. Really appreciate the contribution, but I don't really like the API. It's a tough one though,  I'll think about it some more and try to propose an alternative some time this week.

maleadt:
Superseded by https://github.com/JuliaGPU/CUDAdrv.jl/pull/133

PR #125, opened 2019-02-03
WIP/RFC: Introduce a CuPtr type to avoid erroneous CPU/GPU conversions.

maleadt:
As suggested by @Keno, this introduces a separate pointer type for device pointers in order to avoid accidental conversions like we've seen a couple over at CuArrays.jl and Flux.jl (eg. https://github.com/FluxML/Flux.jl/issues/581 and https://github.com/JuliaGPU/CuArrays.jl/issues/264).

This change will need downstream adjustments to at least CUDAnative and CuArrays, and will break user code that does manual `cudacall(..., Ptr, ...)`, so I'd like some thoughts here before pushing this.

One open question is how to deal with the fact that unified/managed pointers are both accessible from the host and the device, ie. they are both `Ptr` and `CuPtr`. ~I considered adding a type attribute that captures the pointer type (host, device, or managed) but that still wouldn't allow reuse of existing code (eg. `unsafe_wrap(Array{T}, ::CuPtr{T})`) and function signatures would be pretty messy (eg. `cublas_call(::CuPtr{T,P} where P <: Union{Device,Unified})`). Maybe `CuPtr` should just represent a device pointer, regular or managed, with some very explicit convert method to go from `CuPtr` to `Ptr`.~
EDIT: implemented something like that, type aliases should help.


maleadt:
OK, added the address space to the pointer type. Not sure if it's worth the churn, but at least it lets us safely convert to regular `Ptr`.

vchuravy:
The AS makes sense to me. It does get a bit weird on systems with HMM [1] since afaik any allocation will be valid across device boundaries (see the first example in [2]).
I hadn't seen a system that support HMM, but according to the docs on Summit (Power 9 + V100), does support arbitrary allocations [3]. So I think the best thing is to differentiate between allocations
that come from the CUDA driver, but also have a environment flag or something similar that allows for just using a `Ptr`?


- [1] https://www.kernel.org/doc/html/v4.18/vm/hmm.html
- [2] https://www.redhat.com/files/summit/session-assets/2017/S104078-hubbard.pdf
- [3] https://www.olcf.ornl.gov/wp-content/uploads/2018/03/ORNL_workshop_mar2018.pdf

MikeInnes:
Having CuPtr certainly seems like a good idea. Rather than address spaces, would it make sense for `pointer` to generally just return either `CuPtr` or `Ptr` depending on whether access on the host is valid?

maleadt:
I think that would be dangerous, since only managed pointers (and not all <:Ptr) can be used with certain CUDA API calls.

MikeInnes:
Isn't it still a strict improvement over the current situation? Or are you saying that the current situation is dangerous?

(FWIW I said `pointer` earlier but didn't specifically mean that function -- just saw Keno's comment about ccall.)

maleadt:
> Isn't it still a strict improvement over the current situation?

It's still a strict improvement yes, but I'd like to deal with that other ambiguity too now that we're at it. The idea is that `CuPtr` represents everything allocated by CUDA, making it possible to accurately dispatch on that. CUDAnative changes at https://github.com/JuliaGPU/CUDAnative.jl/pull/337/files

maleadt:
Also to bikeshed: should this throw `InexactError`, or just `error`?

KristofferC:
Imo, just `error` (or a dedicated exception for this specific error).

vchuravy:
> Also to bikeshed: should this throw InexactError, or just error?

I would say `ArgumentError`, but just error is also fine.

maleadt:
After thinking about it a little more, I ditched the address-space (KISS). The idea is now that Ptr represents addresses that can be resolved on the CPU, while CuPtr covers everything that can be accessed from the GPU. That doesn't cover unified memory, and host-allocations (cuHostMalloc -> Ptr) aren't "protected" by the type system (eg. call cuHostFree on a non-CUDA alloc'd pointer), but that's similar to Base.

A remaining wart is that many CUDA APIs accept pointers to either host or device memory (unrelated to unified memory). Now that we've separated Ptr from CuPtr, that is problematic. Current solution is another Ptr type that does a hacky cconvert/unsafe_convert: https://github.com/JuliaGPU/CUDAdrv.jl/pull/125/commits/6d98d607e5ecb2b73a2a63b5e04853c6ea5c8339 (Union types can't be used here, the ccall argument types need to be concrete). Suggestions are appreciated.

PR #130, opened 2019-03-13
implement wait on stream

vchuravy:
CC @keno

### DIFF
--- a/src/events.jl
+++ b/src/events.jl
@@ -62,6 +62,14 @@ Waits for an event to complete.
 synchronize(e::CuEvent) = @apicall(:cuEventSynchronize, (CuEvent_t,), e)
 
 """
+    wait(e::CuEvent, stream=CuDefaultStream())
+
+Make a `stream` wait on a event.
+"""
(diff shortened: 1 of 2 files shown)


vchuravy:
@maleadt any qualms with extending `Base.wait`?

maleadt:
> @maleadt any qualms with extending `Base.wait`?

I'm not sure. Semantics of `Base.wait` are generally a blocking wait on the host, so that would need to be `cuEventSynchronize`. But the wait vs synchronize terminology is a choice by CUDA, so it would be confusing to switch that here (with `Base.wait(::CuEvent) = cuEventSynchronize`). So I guess it's fine?

maleadt:
Yeah I'm not ready to commit to this API, so I removed the extension of `Base.wait`. Specifically, I think it would make more sense to have `Base.wait(::CuEvent)` and `Base.wait(::CuStream)` do the expected thing that is now implemented by `synchronize`, i.e. have the host wait. But then the question is how to make a stream wait; maybe `wait!(::CuStream, ::CuEvent)` (but then we always need to pass a stream)?

vchuravy:
Seems okay, thanks for merging.

PR #134, opened 2019-04-02
Cooperative Kernel Launch

qin-yu:
Now functions in src/execution.jl accepts a `cg::Bool` argument that will allow `_launch()` to call the cooperative kernel launch if set to `true`. Note that some other functions depend on launch may break, for example, the test for `launch()`, which is modified accordingly.

### DIFF
--- a/src/execution.jl
+++ b/src/execution.jl
@@ -35,8 +35,7 @@ const CuDim = Union{Integer,
                     Tuple{Integer, Integer},
                     Tuple{Integer, Integer, Integer}}
 
-# we need a generated function to get an args array,
-# without having to inspect the types at runtime
+# pack arguments in a buffer that CUDA expects
 @generated function pack_arguments(f::Function, args...)
     all(isbitstype, args) || throw(ArgumentError("Arguments to kernel should be bitstype."))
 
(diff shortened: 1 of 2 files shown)


qin-yu:
To avoid breaking existing codes that uses `launch()` directly, should I do:
```
@inline function launch(f::CuFunction, blocks::CuDim, threads::CuDim,
                        shmem::Int, stream::CuStream,
                        args...) = launch(f, false, blocks, threads, shmem, stream, args...)
```

vchuravy:
bors try

maleadt:
Since we're breaking that API, I think I'll layer a kwarg-based redesign on top of this PR. The current API and its implementation is a bit messy, and doing so would make it extensible for the future.
A `cooperative` kwarg would be nicer, too.

maleadt:
> bors try

I'll add bors, too :smile:

qin-yu:
> > bors try
> 
> I'll add bors, too

For the bot to build the [CUDAnative PR](https://github.com/JuliaGPU/CUDAnative.jl/pull/372), I think it needs to build the CUDAdrv PR first and based on that run Bors to build the CUDAnative PR?

maleadt:
bors try

maleadt:
Thanks for the contribution, @qin-yu!

qin-yu:
> Thanks for the contribution, @qin-yu!

Thanks! Without encouragement and (plenty of) help from @maleadt :sunny: and @vchuravy :droplet:, I would not have made any contribution :seedling:

PR #136, opened 2019-04-10
Escape stream in elapsed macro

lcw:
This allows streams not in global scope to be passed into `@elapsed`.


maleadt:
Thanks!
bors r+

PR #137, opened 2019-04-10
Add stream creation flags

lcw:


maleadt:
Looks great! I should probably go through all of the `@enum`s and make them typed using the `@enum Foo::T` syntax, but that doesn't need to be part of this PR.
bors r+

PR #138, opened 2019-04-11
Try to unsafe_convert PtrOrCuPtr in opposite order

bcmichael:
When applicable is false it does more than 40 allocations and allocates several KB. When it is true it does 1 very small allocation. It is also slower when it is false. As far as I can tell PtrOrCuPtr seems to only really be used for passing constants into CUBLAS functions in CuArrays. All of the high level wrappers for CUBLAS expect those constants to be on the CPU so checking the CPU conversion first makes those calls faster and allocate less.

It's not a big difference, but I had some code that made a bunch of BLAS calls and used pretty small arrays to make the tests go quick. In that case it makes a quite noticeable speed difference.

### DIFF
--- a/src/pointer.jl
+++ b/src/pointer.jl
@@ -111,10 +111,10 @@ end
 
 function Base.unsafe_convert(::Type{PtrOrCuPtr{T}}, val) where {T}
     # TODO: this should try/catch since the fallback for `unsafe_convert{Ptr}` calls error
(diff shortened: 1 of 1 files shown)


maleadt:
Makes sense. Did you try against CuArrays tests?

bcmichael:
I have now. It didn't seem to break anything.

PR #157, opened 2019-10-15
Autogenerated wrappers

maleadt:
~TODO: maybe put the generated wrappers in a separate module and export that?~
Too annoying to have to use `CUDAdrv.API.cu...` everywhere. Non-wrapped stuff is prefixes with cu or CU anyway, so that's clear enough methinks.


maleadt:
Great, seems to be non-breaking. Couldn't wrap the entire API though, there's several points where Clang.jl bails out. Its `wrap_c` functionality is pretty hairy, actually, but it didn't seem like CBindingGen.jl did any better so sticking with Clang.jl for now.

PR #172, opened 2019-10-31
Add a mechanism for declaring public but not exported API.

maleadt:


maleadt:
Am probably going to merge the packages instead.

PR #197, opened 2020-01-28
Wrap some simple pointer attribute queries, add tests

kshyatt:
Feel free to edit anything here you don't like! Wasn't sure whether to export these...

### DIFF
--- a/src/pointer.jl
+++ b/src/pointer.jl
@@ -122,15 +122,3 @@ function Base.unsafe_convert(::Type{PtrOrCuPtr{T}}, val) where {T}
     end
     return Base.bitcast(PtrOrCuPtr{T}, ptr)
 end
-
-function memory_type(x::CuPtr)
-    dat = Ref{CUmemorytype}()
-    cuPointerGetAttribute(dat, CU_POINTER_ATTRIBUTE_MEMORY_TYPE, x)
-    return dat[]
-end
-
-function is_managed(x::CuPtr)
-    dat = Ref{UInt32}()
-    cuPointerGetAttribute(dat, CU_POINTER_ATTRIBUTE_IS_MANAGED, x)
-    return convert(Bool, dat[])
-end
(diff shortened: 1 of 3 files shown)


src/pointer.jl
```diff
@@ -122,3 +122,15 @@ function Base.unsafe_convert(::Type{PtrOrCuPtr{T}}, val) where {T}
     end
     return Base.bitcast(PtrOrCuPtr{T}, ptr)
 end
+
+function memory_type(x::CuPtr)
```
maleadt: Maybe export the enum with `@enum_without_prefix` and cast the value to an instance of it?

kshyatt: Sorry, I'm super confused by this comment - are you talking about the return value of this function?

maleadt: Yes, sorry for the dense comment. The return type is now a nondescript integer, right? But we know it should be a value of the CUmemorytype_enum. So it'd be better to return an enum, instead.
However, for users to be able to use that enum, we need to re-export that. That's what the `@enum_without_prefix` macro is for, to remove the redundant CU in `CU_MEMORYTYPE_DEVICE`.

kshyatt:
Moved these to `src/memory.jl` bc `@enum_without_prefix` is defined after `src/pointer.jl` is included. Hope I did this right...

maleadt:
Thanks! I realize I didn't want this exported after all, so I pushed a commit.

I also loosened the type signature because you can then rely on the magic of `cconvert`:

```julia
julia> using CuArrays, CUDAdrv

julia> a = CuArray([1])
1-element CuArray{Int64,1,Nothing}:
 1

julia> is_managed(a)
false

julia> b = unsafe_wrap(CuArray, convert(CuPtr{Int}, Mem.alloc(Mem.Unified, sizeof(Int))), (1,))
1-element CuArray{Int64,1,Nothing}:
 0

julia> is_managed(b)
true
```

EDIT: actually, how nice this may be there's also a `attribute(::CuDevice)` so we need some type information to disambiguate...

maleadt:
Let's keep it at `is_managed(pointer(...))`, that keeps multiple dispatch happy.

Made it work on regular `Ptr` too so you can safely work with unified `Array` types too.

maleadt:
@kshyatt Go ahead and merge if these changes works for you.

kshyatt:
Should be fine, I'll make another PR if there are problems.

PR #202, opened 2020-02-02
add query for stream and event

vchuravy:
While implementing an MPI friendly `synchronize(stream)`, I noticed that the API swallows the error code that query is supposed to return. x-ref: https://github.com/climate-machine/CLIMA/pull/439/commits/a4a6c1610767f61da69564e5a862995ee89368ac#diff-f93e99470aee8fd224be784002b3054dR169


maleadt:
Argh.

PR #204, opened 2020-02-05
Generate unsafe wrappers that return the status instead of throwing.

maleadt:
To make https://github.com/JuliaGPU/CUDAdrv.jl/pull/202 possible without hacking `libcuda.jl`.


vchuravy:
Test failed for me locally before 6bf9aba, escaping the function names should be the right thing here.

```
ERROR: LoadError: LoadError: UndefVarError: cuDeviceTotalMem_v2 not defined
Stacktrace:
 [1] top-level scope at /home/vchuravy/.julia/dev/CUDAdrv/src/libcuda_aliases.jl:1
 [2] include at ./boot.jl:328 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1105
 [4] include at ./Base.jl:31 [inlined]
 [5] include(::String) at /home/vchuravy/.julia/dev/CUDAdrv/src/CUDAdrv.jl:1
 [6] top-level scope at /home/vchuravy/.julia/dev/CUDAdrv/src/CUDAdrv.jl:21
```

maleadt:
Yeah. I also don't really like the `@eval`, but couldn't figure out a different invocation (do we need `:toplevel` here?). EDIT: ah I see you got rid of it.

PR #209, opened 2020-02-17
Fix version check

timholy:
I was getting
```
┌ Warning: CUDAdrv.jl only supports NVIDIA drivers for CUDA 9.0 or higher (yours is for CUDA 9.0.0)
└ @ CUDAdrv ~/.julia/dev/CUDAdrv/src/CUDAdrv.jl:57
```

### DIFF
--- a/src/CUDAdrv.jl
+++ b/src/CUDAdrv.jl
@@ -53,7 +53,7 @@ function __init__()
 
         cuInit(0)
 
-        if version() <= v"9"
+        if version() < v"9"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,2 +1,4 @@
 *.jl.*.cov
 *.jl.mem
+deps/ext.jl
```
maleadt: Where do these come from? We haven't had a `deps` folder in a loong time.

timholy: Dunno, but I dropped it.

maleadt:
Thanks!

PR #230, opened 2020-04-01
add Mem.pin that tracks registered Arrays

vchuravy:
This is needed in KernelAbstractions for `async_copy` since
that needs to be to and from registered memory regions,
and @kshyatt needed something like this for her work as well.

Putting this up so that we can discuss implementation and location of the code.

### DIFF
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -185,7 +185,7 @@ end
 # - Deal with memory regions/views
 const __pinned_memory = Dict{Ptr, WeakRef}()
 
-function pin(a, flags=0)
+function pin(a::Array, flags=0)
     # use pointer instead of objectid?
     oid = pointer(a)
     if haskey(__pinned_memory, oid) && __pinned_memory[oid].value !== nothing
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -185,13 +185,13 @@ end
 # - Deal with memory regions/views
 const __pinned_memory = Dict{Ptr, WeakRef}()
 
-function pin(a)
+function pin(a, flags=0)
     # use pointer instead of objectid?
     oid = pointer(a)
     if haskey(__pinned_memory, oid) && __pinned_memory[oid].value !== nothing
         return nothing
     end
-    ad = Mem.register(Mem.Host, pointer(a), sizeof(a))
+    ad = Mem.register(Mem.Host, pointer(a), sizeof(a), flags)
     finalizer(_ -> Mem.unregister(ad), a)
     __pinned_memory[oid] = WeakRef(a)
     return nothing
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -187,13 +187,13 @@ const __pinned_memory = Dict{Ptr, WeakRef}()
 
 function pin(a::Array, flags=0)
     # use pointer instead of objectid?
-    oid = pointer(a)
-    if haskey(__pinned_memory, oid) && __pinned_memory[oid].value !== nothing
+    ptr = pointer(a)
+    if haskey(__pinned_memory, ptr) && __pinned_memory[ptr].value !== nothing
         return nothing
     end
     ad = Mem.register(Mem.Host, pointer(a), sizeof(a), flags)
     finalizer(_ -> Mem.unregister(ad), a)
-    __pinned_memory[oid] = WeakRef(a)
+    __pinned_memory[ptr] = WeakRef(a)
     return nothing
 end
 
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -183,11 +183,11 @@ end
 # - WeakRef dict does not unique the key by objectid
 # TODO:
 # - Deal with memory regions/views
-const __pinned_memory = Dict{UInt64, WeakRef}()
+const __pinned_memory = Dict{Ptr, WeakRef}()
 
 function pin(a)
     # use pointer instead of objectid?
-    oid = objectid(a)
+    oid = pointer(a)
     if haskey(__pinned_memory, oid) && __pinned_memory[oid].value !== nothing
         return nothing
     end
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -176,6 +176,26 @@ function free(buf::HostBuffer)
     end
 end
 
+##
+# Pinned memory management
+#
+# - IdDict does not free the memory
+# - WeakRef dict does not unique the key by objectid
+# TODO:
+# - Deal with memory regions/views
+const __pinned_memory = Dict{UInt64, WeakRef}()
+
+function pin(a)
+    # use pointer instead of objectid?
+    oid = objectid(a)
+    if haskey(__pinned_memory, oid) && __pinned_memory[oid].value !== nothing
+        return nothing
+    end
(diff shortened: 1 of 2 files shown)


maleadt:
Hm, generally CUDAdrv doesn't have these smarts; there's other API calls that would fail if you call them twice too. Seems like a simple enough tool to consider though.

src/memory.jl
```diff
@@ -176,6 +176,26 @@ function free(buf::HostBuffer)
...
+# - IdDict does not free the memory
+# - WeakRef dict does not unique the key by objectid
+# TODO:
+# - Deal with memory regions/views
+const __pinned_memory = Dict{UInt64, WeakRef}()
+
+function pin(a)
+    # use pointer instead of objectid?
+    oid = objectid(a)
+    if haskey(__pinned_memory, oid) && __pinned_memory[oid].value !== nothing
```
maleadt: Why not `get!` on a `WeakKeyDict`?

vchuravy: WeakKeyDict doesn't use objectid:

```
julia> D = WeakKeyDict()
WeakKeyDict{Any,Any} with 0 entries

julia> A = zeros(3)
3-element Array{Float64,1}:
 0.0
 0.0
 0.0

julia> D[A] = 1
1

julia> A[1] = 1
1

julia> D[A] = 2
2

julia> D
WeakKeyDict{Any,Any} with 2 entries:
  [1.0, 0.0, 0.0] => 1
  [1.0, 0.0, 0.0] => 2
```

kshyatt:
I changed the `Dict` key type to `Ptr` - before the change this wasn't working (extremely slow) in my code and now it seems to be in all my multiGPU tests, if that counts for anything?

kshyatt:
What are our thoughts on this -- it seems we don't have tests? I can add some?

maleadt:
Yes, please. There were also some TODO's in the code, if those aren't important or of the everlasting kind go ahead and merge once we have some test coverage at the least.

kshyatt:
I think the TODO can rest for a while... These tests are pretty basic, just making sure that we haven't totally destroyed the device memory.

kshyatt:
OK, the tests passed but there's some of bad token issue or something

kshyatt:
Added a way to pass flags so you can explicitly device map the buffer

src/memory.jl
```diff
@@ -176,6 +176,26 @@ function free(buf::HostBuffer)
...
 end
 
+##
+# Pinned memory management
+#
+# - IdDict does not free the memory
+# - WeakRef dict does not unique the key by objectid
+# TODO:
+# - Deal with memory regions/views
+const __pinned_memory = Dict{Ptr, WeakRef}()
```
vchuravy, line 186: @kshyatt I think with using a `Ptr` we no longer need the `WeakRef`.

src/memory.jl
```diff
@@ -176,6 +176,26 @@ function free(buf::HostBuffer)
...
+
+function pin(a, flags=0)
```
vchuravy:
```suggestion
function pin(a::Array, flags=0)
```

since we don't want to deal with views.

src/memory.jl
kshyatt: I'll try later today without it

kshyatt:
@vchuravy suggested changing the `Dict` to a `Set` of pointers, but this lead to finalizer errors in my code whereas the `Dict` version seems to work without problems. @maleadt are you ok with this as-is?

src/memory.jl
```diff
@@ -176,6 +176,26 @@ function free(buf::HostBuffer)
...
+# Pinned memory management
+#
+# - IdDict does not free the memory
+# - WeakRef dict does not unique the key by objectid
+# TODO:
+# - Deal with memory regions/views
+const __pinned_memory = Dict{Ptr, WeakRef}()
+
+function pin(a::Array, flags=0)
+    # use pointer instead of objectid?
```
maleadt, line 189: comment doesn't seem to match code below?

kshyatt: We didn't rename `oid` but we are actually using the pointer

kshyatt: Fixed the naming issue

kshyatt:
Better now?

src/memory.jl
```diff
@@ -176,6 +176,26 @@ function free(buf::HostBuffer)
...
+    ptr = pointer(a)
+    if haskey(__pinned_memory, ptr) && __pinned_memory[ptr].value !== nothing
+        return nothing
```
maleadt, line 192:
```suggestion
        @assert __pinned_memory[ptr].value === a
        return nothing
```

I don't really get this though, so you're not allowed to register memory twice, but wat if a Julia array is freed (`.value === nothing`) but the memory reused; won't you then end up registering the same pointer, now backed by a different array, twice?

vchuravy: Right, Looking at UCX they have an `rcache` which is a cache for memory regions that are registered, and that supports merging of regions.
I was hoping to come up with a less complex implementation, but I am not a 100% how to best deal with `reallocs`, and free's and allocs that are faster than the finalizers. Since @kshyatt reported that the `Set` implementation can cause finalizer errors I suspect we need to be more careful, since it basically means that we can see GC freeing an array an then re-using its memory before we get to run the finalizer.

kshyatt: FWIW using this code "in production" with many thousands of really big arrays (I can try to check my memory footprint) I haven't hit this issue.

maleadt:
IntervalTrees.jl is a pretty lightweight dependency, so we could just use it. It doesn't merge intervals though, so I came up with the following:

```julia
using IntervalTrees

const registered = IntervalTree{Int, Interval{Int}}()

function register(registered, low, high)
    @info "Want to register $low -> $high"

    # register gaps between already-registered slices
    intersections = [Interval{Int}(0, low),
                     intersect(registered, Interval{Int}(low, high))...,
                     Interval{Int}(high, 0)]
    for (lower, upper) in zip(intersections[1:end-1], intersections[2:end])
        gap = Interval{Int}(last(lower), first(upper))
        if first(gap) < last(gap)
            @warn "Actually register $(first(gap)) -> $(last(gap))"
        end
    end

    # delete actual intersections
    low = min(low, first(intersections[2]))
    high = max(high, last(intersections[end-1]))
    for range in intersections[2:end-1]
        @info "Deleting $(first(range)) -> $(last(range))"
        delete!(registered, range)
    end

    @info "Insert $low -> $high"
    push!(registered, Interval{Int}(low, high))
    display(registered)
    println()
    println()
end

register(registered, 100, 200)
register(registered, 50, 250)   # should subsume the previous one
register(registered, 225, 230)  # shouldn't do anything
register(registered, 250, 300)  # should extend
register(registered, 25, 50)    # should extend
register(registered, 0, 100)    # should partially extend
register(registered, 250, 350)  # should partially extend
```

Thoughts?

kshyatt:
@maleadt if you can make a PR with this to CUDA.jl, I can try it with my big giant eighteenwheeler of an app

maleadt:
https://github.com/JuliaGPU/CUDA.jl/pull/233

PR #237, opened 2020-05-01
Wrap functionality to set and query stream priorities.

### LINKED ISSUE #146: Add support for CUDA stream creation with priority
The objective is e.g. to be able to achieve the equivalent of what the following CUDA C code does:
```
cudaStream_t streams[2];
int leastPriority=-1, greatestPriority=-1;
cudaDeviceGetStreamPriorityRange(&leastPriority, &greatestPriority);
cudaStreamCreateWithPriority(&streams[0], cudaStreamNonBlocking, greatestPriority);
cudaStreamCreateWithPriority(&streams[1], cudaStreamNonBlocking, leastPriority);
```
The importance of this feature was shown in [this topic](https://discourse.julialang.org/t/cuda-streams-do-not-overlap/24879/5) in Julia discourse.

maleadt:
Fixes https://github.com/JuliaGPU/CUDAdrv.jl/issues/146

### DIFF
--- a/src/stream.jl
+++ b/src/stream.jl
@@ -1,7 +1,7 @@
 # Stream management
 
 export
-    CuStream, CuDefaultStream, synchronize
+    CuStream, CuDefaultStream, priority, priority_range, synchronize
 
 
 mutable struct CuStream
@@ -17,13 +17,19 @@ Base.hash(s::CuStream, h::UInt) = hash(s.handle, h)
 @enum_without_prefix CUstream_flags_enum CU_
 
 """
-    CuStream(flags=STREAM_DEFAULT)
+    CuStream(; flags=STREAM_DEFAULT, priority=nothing)
 
 Create a CUDA stream.
 """
-function CuStream(flags::CUstream_flags=STREAM_DEFAULT)
+function CuStream(; flags::CUstream_flags=STREAM_DEFAULT,
+                    priority::Union{Nothing,Integer}=nothing)
     handle_ref = Ref{CUstream}()
-    cuStreamCreate(handle_ref, flags)
+    if priority === nothing
+        cuStreamCreate(handle_ref, flags)
+    else
+        priority in priority_range() || throw(ArgumentError("Priority is out of range"))
+        cuStreamCreateWithPriority(handle_ref, flags, priority)
+    end
 
     ctx = CuCurrentContext()
     obj = CuStream(handle_ref[], ctx)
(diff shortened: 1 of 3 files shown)


maleadt:
cc @samo-lin, could you have a look?

src/stream.jl
```diff
@@ -17,13 +17,19 @@ Base.hash(s::CuStream, h::UInt) = hash(s.handle, h)
...
 
 """
-    CuStream(flags=STREAM_DEFAULT)
+    CuStream(; flags=STREAM_DEFAULT, priority=nothing)
 
 Create a CUDA stream.
 """
-function CuStream(flags::CUstream_flags=STREAM_DEFAULT)
+function CuStream(; flags::CUstream_flags=STREAM_DEFAULT,
+                    priority::Union{Nothing,Int}=nothing)
```
samo-lin: In my simple test, where I gave priorities obtained from `priority_range()` to `CuStream` it failed with this error:
```
omlins@nid00003:~/.julia/1.4.1/dom-gpu/dev/CUDAdrv> julia
    
julia> using CUDAdrv, CUDAnative, CuArrays

julia> p_range = priority_range();


julia> s1 = CuStream(; flags=CUDAdrv.STREAM_NON_BLOCKING, priority=minimum(p_range))
ERROR: TypeError: in keyword argument priority, expected Union{Nothing, Int64}, got Int32
Stacktrace:
 [1] top-level scope at REPL[3]:1
```
I think the elements obtained from the output of `priority_range()` should not require any type conversion when passing them to `CuStream`. Maybe  to do just `priority::Union{Nothing,Integer}=nothing` would be fine (after I changed this the CuStream was successfully created)?

samo-lin (changes requested):
Once this little fix of the argument type of CuStream is done it seems all perfectly fine. Using my local fix, I could successfully overlap two streams using the following code (overlap verified with nvvp):
```
using CUDAdrv, CUDAnative, CuArrays

function memcopy!(A, B)
    ix = (blockIdx().x-1) * blockDim().x + threadIdx().x
    A[ix] = B[ix]
    return
end

nx = 128*1024^2
nt = 100
A = CuArrays.zeros(nx);
B = CuArrays.ones(nx);
C = CuArrays.zeros(nx);
D = CuArrays.ones(nx);
nthreads = 1024
nblocks = ceil(Int, nx/nthreads)

p_range = priority_range();
s1 = CuStream(; flags=CUDAdrv.STREAM_NON_BLOCKING, priority=p_range[1])
s2 = CuStream(; flags=CUDAdrv.STREAM_NON_BLOCKING, priority=p_range[end])

priority(s1)
priority(s2)

for it = 1:nt
    @cuda blocks=nblocks threads=nthreads stream=s1 memcopy!(A, B);
    @cuda blocks=nblocks threads=nthreads stream=s2 memcopy!(C, D);
    CUDAdrv.synchronize()
end
```

src/stream.jl
maleadt: Yes, that makes sense. Thanks for looking it over!

samo-lin:
Thanks @maleadt !!

PR #238, opened 2020-05-01
Wrap 3d memcpy

### LINKED ISSUE #149: Add support for asynchronous 3-D memcopy (cudaMemcpy3DAsync)
The objective is e.g. to be able to achieve the equivalent of what the following CUDA C code does:
```
int main(){
  const size_t nx = 512;
  const size_t ny = 512;
  const size_t nz = 512;
  float *d_src, *h_dst;
  cudaMalloc(&d_src, nx*ny*nz*sizeof(float));
  cudaHostAlloc(&h_dst, nx*nz*sizeof(float), cudaHostAllocDefault);
  cudaMemset(d_src, 0, nx*ny*nz*sizeof(float));
  memset(h_dst, 1, nx*nz*sizeof(float));

  cudaMemcpy3DParms cpy_params = {0};
  cpy_params.srcPtr = make_cudaPitchedPtr(d_src, nx*sizeof(float), nx, ny);
  cpy_params.dstPtr = make_cudaPitchedPtr(h_dst, nx*sizeof(float), nx, 1);
  cpy_params.srcPos = make_cudaPos((size_t) 0, (size_t) 1, (size_t) 0);
  cpy_params.dstPos = make_cudaPos((size_t) 0, (size_t) 0, (size_t) 0);
  cpy_params.extent = make_cudaExtent(nx*sizeof(float), 1, nz);
  cpy_params.kind   = cudaMemcpyDeviceToHost;

  for (int i=0; i<100; i++){
    cudaMemcpy3DAsync(&cpy_params);
  }

  cudaDeviceSynchronize();
}
```
It reaches about 12 GB/s (Giga: 1e9) with 16 PCIe 3.0 lanes and a NVIDIA Tesla P100.

A corresponding code using multiple 1-D memcopy calls instead reached only about 1.5 GB/s in some experiments reported in the Julia discourse. The Julia discourse topic discussing this issue is found [here](https://discourse.julialang.org/t/cuarrays-cudanative-cudadrv-equivalent-to-cudamemcpy3dasync/26182/8).

maleadt:
Fixes https://github.com/JuliaGPU/CUDAdrv.jl/issues/149

### DIFF
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -327,12 +327,23 @@ for (f, srcPtrTy, dstPtrTy) in (("cuMemcpyDtoH", CuPtr, Ptr),
     end
 end
 
+"""
+    unsafe_copy3d!(dst, dstTyp, src, srcTyp, width, height=1, depth=1;
+                   dstPos=(1,1,1), dstPitch=0, dstHeight=0,
+                   srcPos=(1,1,1), srcPitch=0, srcHeight=0,
+                   async=false, stream=nothing)
+
+Perform a 3D memory copy between pointers `src` and `dst`, at respectively position `srcPos`
+and `dstPos` (1-indexed). Both pitch and destination can be specified for both the source
+and destination; consult the CUDA documentation for more details. This call is executed
+asynchronously if `async` is set, in which case `stream` needs to be a valid CuStream.
+"""
 function unsafe_copy3d!(dst::Union{Ptr{T},CuPtr{T}}, dstTyp::Type{<:Buffer},
                         src::Union{Ptr{T},CuPtr{T}}, srcTyp::Type{<:Buffer},
                         width::Integer, height::Integer=1, depth::Integer=1;
-                        srcPos::CuDim=(1,1,1), dstPos::CuDim=(1,1,1),
-                        srcPitch::Integer=0, srcHeight::Integer=0,
+                        dstPos::CuDim=(1,1,1), srcPos::CuDim=(1,1,1),
                         dstPitch::Integer=0, dstHeight::Integer=0,
+                        srcPitch::Integer=0, srcHeight::Integer=0,
                         async::Bool=false, stream::Union{Nothing,CuStream}=nothing) where T
     srcPos = CUDAdrv.CuDim3(srcPos)
     dstPos = CUDAdrv.CuDim3(dstPos)
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -5,7 +5,7 @@ export Mem, attribute, attribute!, memory_type, is_managed
 module Mem
 
 using ..CUDAdrv
-using ..CUDAdrv: @enum_without_prefix, CUstream, CUdevice
+using ..CUDAdrv: @enum_without_prefix, CUstream, CUdevice, CuDim3
 
 using Base: @deprecate_binding
 
@@ -324,6 +324,70 @@ for (f, srcPtrTy, dstPtrTy) in (("cuMemcpyDtoH", CuPtr, Ptr),
             $(getproperty(CUDAdrv, Symbol(f)))(dst, src, N*sizeof(T))
         end
         return dst
+    end
+end
+
+function copy_3d(src::Union{Ptr,CuPtr}, srcTyp::Type{<:Buffer},
+                 dst::Union{Ptr,CuPtr}, dstTyp::Type{<:Buffer},
+                 width::Integer, height::Integer=1, depth::Integer=1;
+                 srcPos::CuDim3=CuDim3(0,0,0), dstPos::CuDim3=CuDim3(0,0,0),
+                 srcPitch::Integer=0, srcHeight::Integer=0,
+                 dstPitch::Integer=0, dstHeight::Integer=0,
+                 async::Bool=false, stream::Union{Nothing,CuStream}=nothing)
+    srcMemoryType, srcHost, srcDevice = if srcTyp == Host
+        CUDAdrv.CU_MEMORYTYPE_HOST,
+        src::Ptr,
+        CU_NULL
+    elseif srcTyp == Mem.Device
+        CUDAdrv.CU_MEMORYTYPE_DEVICE,
+        C_NULL,
+        src::CuPtr
+    elseif srcTyp == Mem.Unified
+        CUDAdrv.CU_MEMORYTYPE_UNIFIED,
+        C_NULL,
+        reinterpret(CuPtr{Cvoid}, src)
+    end
+    srcArray = C_NULL
+
+    dstMemoryType, dstHost, dstDevice = if dstTyp == Host
+        CUDAdrv.CU_MEMORYTYPE_HOST,
+        dst::Ptr,
+        CU_NULL
+    elseif dstTyp == Mem.Device
+        CUDAdrv.CU_MEMORYTYPE_DEVICE,
+        C_NULL,
+        dst::CuPtr
+    elseif dstTyp == Mem.Unified
+        CUDAdrv.CU_MEMORYTYPE_UNIFIED,
+        C_NULL,
+        reinterpret(CuPtr{Cvoid}, dst)
+    end
+    dstArray = C_NULL
+
+    params_ref = Ref(CUDAdrv.CUDA_MEMCPY3D(
+        # source
+        srcPos.x, srcPos.y, srcPos.z,
+        0, # LOD
+        srcMemoryType, srcHost, srcDevice, srcArray,
+        C_NULL, # reserved
+        srcPitch, srcHeight,
+        # destination
+        dstPos.x, dstPos.y, dstPos.z,
+        0, # LOD
+        dstMemoryType, dstHost, dstDevice, dstArray,
+        C_NULL, # reserved
+        dstPitch, dstHeight,
+        # extent
+        width, height, depth
+    ))
+    if async
+        stream===nothing &&
+            throw(ArgumentError("Asynchronous memory operations require a stream."))
+        CUDAdrv.cuMemcpy3DAsync_v2(params_ref, stream)
+    else
+        stream===nothing ||
+            throw(ArgumentError("Synchronous memory operations cannot be issued on a stream."))
+        CUDAdrv.cuMemcpy3D_v2(params_ref)
     end
 end
 
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -327,13 +327,13 @@ for (f, srcPtrTy, dstPtrTy) in (("cuMemcpyDtoH", CuPtr, Ptr),
     end
 end
 
-function copy_3d(src::Union{Ptr,CuPtr}, srcTyp::Type{<:Buffer},
-                 dst::Union{Ptr,CuPtr}, dstTyp::Type{<:Buffer},
-                 width::Integer, height::Integer=1, depth::Integer=1;
-                 srcPos::CuDim3=CuDim3(0,0,0), dstPos::CuDim3=CuDim3(0,0,0),
-                 srcPitch::Integer=0, srcHeight::Integer=0,
-                 dstPitch::Integer=0, dstHeight::Integer=0,
-                 async::Bool=false, stream::Union{Nothing,CuStream}=nothing)
+function unsafe_copy3d!(dst::Union{Ptr{T},CuPtr{T}}, dstTyp::Type{<:Buffer},
+                        src::Union{Ptr{T},CuPtr{T}}, srcTyp::Type{<:Buffer},
+                        width::Integer, height::Integer=1, depth::Integer=1;
+                        srcPos::CuDim3=CuDim3(0,0,0), dstPos::CuDim3=CuDim3(0,0,0),
+                        srcPitch::Integer=0, srcHeight::Integer=0,
+                        dstPitch::Integer=0, dstHeight::Integer=0,
+                        async::Bool=false, stream::Union{Nothing,CuStream}=nothing) where T
     srcMemoryType, srcHost, srcDevice = if srcTyp == Host
         CUDAdrv.CU_MEMORYTYPE_HOST,
         src::Ptr,
@@ -378,7 +378,7 @@ function copy_3d(src::Union{Ptr,CuPtr}, srcTyp::Type{<:Buffer},
         C_NULL, # reserved
         dstPitch, dstHeight,
         # extent
-        width, height, depth
+        width*sizeof(T), height, depth
     ))
     if async
         stream===nothing &&
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -330,10 +330,13 @@ end
 function unsafe_copy3d!(dst::Union{Ptr{T},CuPtr{T}}, dstTyp::Type{<:Buffer},
                         src::Union{Ptr{T},CuPtr{T}}, srcTyp::Type{<:Buffer},
                         width::Integer, height::Integer=1, depth::Integer=1;
-                        srcPos::CuDim3=CuDim3(0,0,0), dstPos::CuDim3=CuDim3(0,0,0),
+                        srcPos::CuDim=(1,1,1), dstPos::CuDim=(1,1,1),
                         srcPitch::Integer=0, srcHeight::Integer=0,
                         dstPitch::Integer=0, dstHeight::Integer=0,
                         async::Bool=false, stream::Union{Nothing,CuStream}=nothing) where T
+    srcPos = CUDAdrv.CuDim3(srcPos)
+    dstPos = CUDAdrv.CuDim3(dstPos)
+
     srcMemoryType, srcHost, srcDevice = if srcTyp == Host
         CUDAdrv.CU_MEMORYTYPE_HOST,
         src::Ptr,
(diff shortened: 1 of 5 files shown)


maleadt:
cc @samo-lin, could you have a look? Could use a better test case to actually perform a 3D copy. Thoughts on a more high-level API are welcome too.

samo-lin:
@maleadt , thanks for this already!

> Thoughts on a more high-level API are welcome too.

What do you think about a CUDAnative function with a signature approximately like the following?
```
copy3d!(
src::AbstractArray{}, 
srcranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}, 
dst::AbstractArray{}, 
dstranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}; 
async::Bool=false, 
stream::Union{Nothing,CuStream}=nothing
)
```

Moreover, CuArrays could then maybe invoke this function automatically for 3-D copies in vectorized notation (e.g. A[:, :] = B[:, 2, :]) that match the requirements.

> Could use a better test case to actually perform a 3D copy.

I will test the current function and come up with a suggestion.

src/memory.jl
```diff
@@ -327,6 +327,70 @@ for (f, srcPtrTy, dstPtrTy) in (("cuMemcpyDtoH", CuPtr, Ptr),
     end
 end
 
+function copy_3d(src::Union{Ptr,CuPtr}, srcTyp::Type{<:Buffer},
+                 dst::Union{Ptr,CuPtr}, dstTyp::Type{<:Buffer},
+                 width::Integer, height::Integer=1, depth::Integer=1;
```
samo-lin: `width` is in bytes while `height` and `depth` are just number of elements, as in the draft in the corresponding issue #149, right? If so, I think it is clearer to call it `widthInBytes` as there. If it is called just `width`, then I would expect that it is a number of elements, like `height` and `depth`, which is then automatically multiplied with the size of an element when passed further to the underlying CUDA call. 
Even if this `widthInBytes` is maybe not the nicest thing to have, I believe it is better if the arguments have the same meaning as in the corresponding [CUDA C function](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g79f4f3fde6ae0f529568d881d9e11987). This will probably prevent some bugs when people translate from CUDA C to Julia using CUDAdrv...

maleadt: I was mimicking the CUDA runtime API, which is supposedly what people will be porting from: http://developer.download.nvidia.com/compute/cuda/2_3/toolkit/docs/online/group__CUDART__MEMORY_gc1372614eb614f4689fbb82b4692d30a.html
There, the elements (when using pointers) are `unsigned char` aka bytes, so it seemed reasonable to just call it `width`?

maleadt:
> What do you think about a CUDAnative function with a signature approximately like the following?

The problem is that we can't know from an array if it's host|device, or unified memory.

samo-lin:
I tested the performance with the following code which copies an x-z plane from a 3-D array at position (0, 1, 0) to a 2-D array on the host and back. It corresponds to the previously used code for PCIe 3.0 transfer performance measures, including the CUDA C code in #149.
```
jl_CUDAdrv6> cat h_d_transfers_3d_cpy3d.jl
using CUDAdrv, CUDAnative, CuArrays

warmup = 3
nt = 10*512
nx, ny, nz = 512, 512, 512
A = zeros(Float64, ny, nz);
B = CuArrays.rand(Float64, nx, ny, nz);
Mem.register(Mem.Host, pointer(A), sizeof(A))
custream = CuStream(CUDAdrv.STREAM_NON_BLOCKING);

# Copy from device to host.
for it = 1:nt+warmup
    if (it == warmup+1) global t0 = time() end
    Mem.copy_3d(pointer(B), Mem.Device, pointer(A), Mem.Host, nx*sizeof(A[1]), 1, nz; srcPos=CUDAdrv.CuDim3(0,1,0), srcPitch=nx*sizeof(A[1]), srcHeight=ny, dstPitch=nx*sizeof(A[1]), dstHeight=1, async=true, stream=custream)
    CUDAdrv.synchronize();
end
time_s = time() - t0;  
ntransfers = 1  #Number of host-device transfers per iteration
GBs = 1.0/1e9*nt*ny*nz*sizeof(Float64)*ntransfers/time_s;
println("d2h: time: $time_s; GB/s: $GBs")

# Copy from host to device.
for it = 1:nt+warmup
    if (it == warmup+1) global t0 = time() end
    Mem.copy_3d(pointer(A), Mem.Host, pointer(B), Mem.Device, nx*sizeof(A[1]), 1, nz; dstPos=CUDAdrv.CuDim3(0,1,0), srcPitch=nx*sizeof(A[1]), srcHeight=1, dstPitch=nx*sizeof(A[1]), dstHeight=ny, async=true, stream=custream)
    CUDAdrv.synchronize();
end
time_s = time() - t0;  
ntransfers = 1  #Number of host-device transfers per iteration
GBs = 1.0/1e9*nt*ny*nz*sizeof(Float64)*ntransfers/time_s;
println("h2d: time: $time_s; GB/s: $GBs")
```

The obtained performance is (1G = 1e9 as typically used by vendors when announcing peak performance): 
```
jl_CUDAdrv6> jul -O3 --check-bounds=no h_d_transfers_3d_cpy3d.jl
jl_CUDAdrv6> jul -O3 --check-bounds=no h_d_transfers_3d_cpy3d.jl
d2h: time: 0.8843080997467041; GB/s: 12.142168824503091
h2d: time: 0.9401581287384033; GB/s: 11.420864120388476
```

Without page locked memory, i.e. without the line `Mem.register(Mem.Host, pointer(A), sizeof(A))` the same code gives the following performance:
```
jl_CUDAdrv6> jul -O3 --check-bounds=no h_d_transfers_3d_cpy3d.jl
> jul -O3 --check-bounds=no h_d_transfers_3d_cpy3d.jl
d2h: time: 1.2458829879760742; GB/s: 8.618319973565768
h2d: time: 1.2862639427185059; GB/s: 8.347756540004204
```

Thus, I would say the achieved performance is perfectly fine!

samo-lin:
> Could use a better test case to actually perform a 3D copy.

How about using the code I used for peformance testing as a basis for the functionality tests? It is a real-world use case (to perform 3-D halo updates), using most of the functionality of the 3D memcopy function.

samo-lin:
> The problem is that we can't know from an array if it's host|device, or unified memory.

I see. Maybe something along the following lines? That is, the idea would be to use multiple dispatch with specialization for CuArray arguments. For CuArrays one could determine if it is a host/device/unifiedmem array; for the AbstractArrays one would assume always that it is a host array.
```
copy3d!(
src::CuArray{}, 
srcranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}, 
dst::CuArray{}, 
dstranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}; 
async::Bool=false, 
stream::Union{Nothing,CuStream}=nothing
)

copy3d!(
src::CuArray{}, 
srcranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}, 
dst::AbstractArray{}, 
dstranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}; 
async::Bool=false, 
stream::Union{Nothing,CuStream}=nothing
)

copy3d!(
src::AbstractArray{}, 
srcranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}, 
dst::CuArray{}, 
dstranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}; 
async::Bool=false, 
stream::Union{Nothing,CuStream}=nothing
)

copy3d!(
src::AbstractArray{}, 
srcranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}, 
dst::AbstractArray{}, 
dstranges::Tuple{Union{UnitRange{},Integer},Union{UnitRange{},Integer},Union{UnitRange{},Integer}}; 
async::Bool=false, 
stream::Union{Nothing,CuStream}=nothing
)
```

src/memory.jl
samo-lin: OK, I understand. It is clearly written in the [cudaExtent Struct Reference](http://developer.download.nvidia.com/compute/cuda/2_3/toolkit/docs/online/structcudaExtent.html) that all `width`, `height` and `depth` are supposed to be given in bytes. However, from my tests and experience, I conclude that this is an error / bad wording in the documentation. I think that `height` and `depth` have to be just the number of elements, not the amount of bytes (i.e. as in the corresponding CUDA C function). In the end, it does not make that much sense to give the second or third dimension in bytes, does it?

If I change for example:
```
Mem.copy_3d(pointer(B), Mem.Device, pointer(A), Mem.Host, nx*sizeof(A[1]), 1, 
nz,...
```
to
```
Mem.copy_3d(pointer(B), Mem.Device, pointer(A), Mem.Host, nx*sizeof(A[1]), 1, 
nz*sizeof(A[1]),...
```
it fails:
```
jl_CUDAdrv6> jul -O3 --check-bounds=no h_d_transfers_3d_cpy3d_allbytes.jl
ERROR: LoadError: CUDA error: invalid argument (code 1, ERROR_INVALID_VALUE)
Stacktrace:
 [1] throw_api_error(::CUDAdrv.cudaError_enum) at /users/omlins/.julia/1.4.1/dom-gpu/dev/CUDAdrv/src/error.jl:106
 [2] macro expansion at /users/omlins/.julia/1.4.1/dom-gpu/dev/CUDAdrv/src/error.jl:113 [inlined]
 [3] cuMemcpy3DAsync_v2(::Base.RefValue{CUDAdrv.CUDA_MEMCPY3D_st}, ::CuStream) at /scratch/snx3000tds/omlins/9_soft_install_julia_1_4/software/Julia/1.4.1-CrayGNU-19.10-cuda-10.1/extensions/packages/CUDAapi/XuSHC/src/call.jl:93
 [4] #copy_3d#9 at /users/omlins/.julia/1.4.1/dom-gpu/dev/CUDAdrv/src/memory.jl:386 [inlined]
 [5] top-level scope at /users/omlins/0_misc/hpc_julia/0_h_d_transfers/jl_CUDAdrv6/h_d_transfers_3d_cpy3d_allbytes.jl:22
 [6] include(::Module, ::String) at ./Base.jl:377
 [7] exec_options(::Base.JLOptions) at ./client.jl:288
 [8] _start() at ./client.jl:484
```
Also, we are using `CUDAdrv.@apicall(:cuMemcpy3D, ...)` since a long time with correct results passing only `width` in bytes and `height` and `depth` as number of elements. This calls the function from the CUDA runtime API, doesn't it?

So, if you agree that is an error / bad wording in the CUDA runtime API documentation, then I would use `widthInBytes` to help limiting the confusion that might appear due to this part of the CUDA runtime API documentation

maleadt:
> For CuArrays one could determine if it is a host/device/unifiedmem array; for the AbstractArrays one would assume always that it is a host array

No, that's not true, because of `unsafe_wrap!`.

I changed the API a little, mimicking `Base.unsafe_copyto!`: typed pointers let us use the element size to compute `widthInBytes`, and I switched source and destination.

I added a 3D copy test, but still don't fully understand the 3D memcpy semantics to come up with a test that uses a proper pitch and height. Your example doesn't seem to perform a valid copy either, I assume you just use it to benchmark. IIUC, we'd also need to use `cuMallocPitch` to get CUDA to choose an ideal pitch, but we can add that later. The current API should give users all the flexibility they need.

samo-lin:
> I changed the API a little, mimicking Base.unsafe_copyto!: typed pointers let us use the element size to compute widthInBytes, and I switched source and destination.

That looks like a good solution to me.

> I added a 3D copy test, but still don't fully understand the 3D memcpy semantics to come up with a test that uses a proper pitch and height. Your example doesn't seem to perform a valid copy either, I assume you just use it to benchmark.

I couldn't see any problem with my example. I removed the code for performance evaluation in it and added some code to check results (marked with a comment # [inserted] ... ). See here the code:
```
using CUDAdrv, CUDAnative, CuArrays, Test

nx, ny, nz = 4, 4, 4
A = zeros(Float64, ny, nz);
B = CuArrays.rand(Float64, nx, ny, nz);
custream = CuStream(CUDAdrv.STREAM_NON_BLOCKING);

# Copy from device to host.
Mem.copy_3d(pointer(B), Mem.Device, pointer(A), Mem.Host, nx*sizeof(A[1]), 1, nz; srcPos=CUDAdrv.CuDim3(0,1,0), srcPitch=nx*sizeof(A[1]), srcHeight=ny, dstPitch=nx*sizeof(A[1]), dstHeight=1, async=true, stream=custream)
CUDAdrv.synchronize();

# [inserted] Check copy
@show A
@test all(B[:,2,:] .== CuArray(A))

# [inserted] Set a different value for A
A .= 7;

# Copy from host to device.
Mem.copy_3d(pointer(A), Mem.Host, pointer(B), Mem.Device, nx*sizeof(A[1]), 1, nz; dstPos=CUDAdrv.CuDim3(0,1,0), srcPitch=nx*sizeof(A[1]), srcHeight=1, dstPitch=nx*sizeof(A[1]), dstHeight=ny, async=true, stream=custream)
CUDAdrv.synchronize();

# [inserted] Check copy
@show B[:,2,:]
@test all(B[:,2,:] .== CuArray(A))
```

and here is the output in the REPL:
```
jl_CUDAdrv6> julia

julia> nx, ny, nz = 4, 4, 4
(4, 4, 4)

julia> A = zeros(Float64, ny, nz);

julia> B = CuArrays.rand(Float64, nx, ny, nz);

julia> custream = CuStream(CUDAdrv.STREAM_NON_BLOCKING);

julia> # Copy from device to host.
       Mem.copy_3d(pointer(B), Mem.Device, pointer(A), Mem.Host, nx*sizeof(A[1]), 1, nz; srcPos=CUDAdrv.CuDim3(0,1,0), srcPitch=nx*sizeof(A[1]), srcHeight=ny, dstPitch=nx*sizeof(A[1]), dstHeight=1, async=true, stream=custream)

julia> CUDAdrv.synchronize();

julia> # [inserted] Check copy
       @show A
A = [0.7441539832634934 0.5109614275986025 0.9433314281062217 0.542658271025779; 0.18313104455831125 0.9477042140053136 0.20451138318639733 0.3249274193957587; 0.146589441085681 0.7077050480032119 0.0503775745404858 0.9722999066935383; 0.4230956996088489 0.2951648161058726 0.33984290259557753 0.9141875459517013]
4×4 Array{Float64,2}:
 0.744154  0.510961  0.943331   0.542658
 0.183131  0.947704  0.204511   0.324927
 0.146589  0.707705  0.0503776  0.9723
 0.423096  0.295165  0.339843   0.914188

julia> @test all(B[:,2,:] .== CuArray(A))
Test Passed

julia> # [inserted] Set a different value for A
       A .= 7;

julia> # Copy from host to device.
       Mem.copy_3d(pointer(A), Mem.Host, pointer(B), Mem.Device, nx*sizeof(A[1]), 1, nz; dstPos=CUDAdrv.CuDim3(0,1,0), srcPitch=nx*sizeof(A[1]), srcHeight=1, dstPitch=nx*sizeof(A[1]), dstHeight=ny, async=true, stream=custream)

julia> CUDAdrv.synchronize();

julia> # [inserted] Check copy
       @show B[:,2,:]
B[:, 2, :] = [7.0 7.0 7.0 7.0; 7.0 7.0 7.0 7.0; 7.0 7.0 7.0 7.0; 7.0 7.0 7.0 7.0]
4×4 CuArray{Float64,2,Nothing}:
 7.0  7.0  7.0  7.0
 7.0  7.0  7.0  7.0
 7.0  7.0  7.0  7.0
 7.0  7.0  7.0  7.0

julia> @test all(B[:,2,:] .== CuArray(A))
Test Passed

```

What maybe hasn't been clear to you is that the `dstPos` and `srcPos` have to be given as zero-based indices (e.g. `CUDAdrv.CuDim3(0,1,0)` to refer to position (1,2,1) of the array in one-based Julia indices). It should probably be taken care of by the function to convert one-based indices to zero-based indices...

maleadt:
I was using the default indices, so that should have worked. Anyhow, I'll make it 1-based since that's definitely more user friendly. Unless you have any more comments, I'd merge this and tag a release.

src/memory.jl
```diff
@@ -330,10 +330,13 @@ end
 function unsafe_copy3d!(dst::Union{Ptr{T},CuPtr{T}}, dstTyp::Type{<:Buffer},
                         src::Union{Ptr{T},CuPtr{T}}, srcTyp::Type{<:Buffer},
                         width::Integer, height::Integer=1, depth::Integer=1;
-                        srcPos::CuDim3=CuDim3(0,0,0), dstPos::CuDim3=CuDim3(0,0,0),
+                        srcPos::CuDim=(1,1,1), dstPos::CuDim=(1,1,1),
```
samo-lin: Shouln't the default value be CuDim(1,1,1) instead of (1,1,1)?

test/memory.jl
```diff
@@ -174,4 +174,38 @@ let
...
+    Mem.unsafe_copy3d!(typed_pointer(dst, Int), Mem.Device, pointer(data), Mem.Host, 3, 3, 3)
+
+    check = zeros(Int, size(data))
+    Mem.unsafe_copy3d!(pointer(check), Mem.Host, typed_pointer(dst, Int), Mem.Device, length(data))
+
+    @test check == data
+
+    Mem.free(dst)
+end
+
```
samo-lin, line 231: How about adding here now a test copying an x-z plane of a 3-D array?

src/memory.jl
maleadt: CuDim is just an alias for either an integer, or a tuple or 1, 2 or 3 items.

samo-lin (changes requested):
I think after this it is perfectly fine. Thanks a lot!!

samo-lin:
> Unless you have any more comments, I'd merge this and tag a release.
Before you tag a new release, could we maybe check if there is missing something to add for registering and the like (3rd functionality I had mentioned)?

maleadt:
> Before you tag a new release, could we maybe check if there is missing something to add for registering and the like (3rd functionality I had mentioned)?

That's all there, right? https://github.com/JuliaGPU/CUDAdrv.jl/blob/f02dcf55a4c38d8e0bafb4923b0daa8230f0d599/src/memory.jl#L151-L181

Tagging versions is easy enough though, so we shouldn't hold up tagging because of that.

samo-lin:
> That's all there, right?

OK, great, thanks!

PR #243, opened 2020-05-23
More tests for pointers

kshyatt:
A few more [missing tests](https://codecov.io/gh/JuliaGPU/CUDAdrv.jl/src/master/src/pointer.jl)


kshyatt:
BOOM over 90% coverage

